Issue #38, opened 2015-11-04
Branch structure need to be clarified in Contributing.md


vors:
Currently I see 'master' branch and 'live' branch.
It's worth to know branch strategy: where send PRs, when something goes live, etc.

joeyaiello:
Agreed. The only reference I currently have in there is this, but it's definitely not enough:
> When you're ready to contribute your documentation, submit a pull request to the *master* branch

We should get something in there ASAP. For posterity, `master` is the current working branch and `live` is a protected branch that gets published into MSDN. 

Also, this isn't set in stone. `live` has to remain the same, but we can have any number of arbitrarily-named branches for separate experiments/projects (and, internally, we can see this content rendered in staging for each branch). 

Does anyone have any opinions on branching?

kilasuit:
Just as a suggestive 

Why not rename Live to Master & Master to Dev to conform with other Microsoft Repos as well? examples being the OfficeDevPnP Repos at https://github.com/OfficeDev/

This has become a "typical" naming convention and really helps to reduce potential confusion caused by what is current and what isnt.

Or perhaps just rename Master to Dev??

But as I said just a suggestive. 

:-)

DonGill:
I don't this has been resolved yet. @joeyaiello can we sync? We need to lock down on a branching strategy.

juanpablojofre:
Closing as resolved/fixed.
Master branch has been renamed Staging.
Live branch remains the same.

PR #39, opened 2015-11-04
Update configData.md

PerOve:
Missing Role in nodename VM-1

### DIFF
--- a/dsc/configData.md
+++ b/dsc/configData.md
@@ -55,7 +55,7 @@ $MyData =
     @(
         @{
             NodeName = "VM-1";
-            
+            Role     = "WebServer"
         },
 
  


joeyaiello:
Hi @PerOve,

This looks great! Thanks so much for being our first external contributor. 

If I could ask one favor: would you mind basing your pull request against PowerShell:master instead of PowerShell:live. Per #38, we need to be a lot more explicit about our branching strategy, and it looks like we may have a bug where our check fails on a PR to master (which should definitely be fixed). However, we still need to propagate changes through a staging branch before going to live. 

Thanks,
Joey

eslesar:
Addressed with #146

Issue #87, opened 2015-12-18
PowerShell-Docs / wmf / class_newtype.md


GoateePFE:
Code samples do not display correctly on the production web site.

joeyaiello:
Yeah, that's pretty egregious. I'm owning this one. Halfway through a fix.

HemantMahawar:
@GoateePFE This issue has been fixed with @joeyaiello PR #114

PR #114, opened 2015-12-31
Fix broken output/links for encrypted MOF, classes, and class-based resources

joeyaiello:


joeyaiello:
As @HemantMahawar pointed out, class_newtype.md really belongs across multiple documents, but I chose to fix the output before refactoring the content.

joeyaiello:
Also, this should fix #87

PR #167, opened 2016-02-17
Removing reference to missing file to unblock publishing to MSDN

HemantMahawar:


HemantMahawar:
@eslesar .. The referenced file complianceServer.md is missing from the repo and publishing to MSDN is broken. I have removed the reference, please add it back once you add the missing file.
@joeyaiello and @neemas  as FYI

PR #215, opened 2016-03-02
Add information about DSC class resources in different .psm1 nested modules

vors:


HemantMahawar:
@vors  Thanks for your updates

PR #295, opened 2016-04-12
no support for SMA 2012 R2

SubZer0MS:
As we don't support WFM 5.0 with SMA 2012 R2, we need to document this as we already found 2 incompatibility bugs and customers should not upgrade to WFM 5.0 anymore in the SMA 2012 R2 environment.

---
This change is [Reviewable](https://reviewable.io/reviews/powershell/powershell-docs/295)


eslesar:
@SubZer0MS Thanks!

PR #386, opened 2016-05-20
Update What-s-New-in-Windows-PowerShell-50.md

CodeCyclone:
Noticed there was some weird characters in front of registered trademarks and the link was still pointing to the Preview build. Not sure if the link needs to be updated to the forwarding link or keep with what I modified it to.

---
This change is [Reviewable](https://reviewable.io/reviews/powershell/powershell-docs/386)


juanpablojofre:
Awesome catch, @CodeCyclone !!

PR #410, opened 2016-05-31
OP provisioning a new docset

saldana:
This change is [Reviewable](https://reviewable.io/reviews/powershell/powershell-docs/410)

### DIFF
--- a/.openpublishing.publish.config.json
+++ b/.openpublishing.publish.config.json
@@ -32,6 +32,16 @@
             "type_mapping": {
                 "Conceptual": "Content"
             }
+        },
+        {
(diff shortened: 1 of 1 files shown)


saldana:
Build will fail because I did not create the JSA folder and I did not add the docfx.json file in the live branch. Those are not needed for provisioning, so I would rather keep your branch clean. Thanks.

eslesar:
Sandra--the live branch is configured to allow a merge only if the build succeeds. How should we proceed here?

saldana:
I will send another PR with all the files so validations pass. Thanks for letting me know.

PR #425, opened 2016-06-14
Moving resource direct call and debugging one level up

HemantMahawar:
As they are applicable to both built-in and custom resources. @eslesar do you agree?

---
This change is [Reviewable](https://reviewable.io/reviews/powershell/powershell-docs/425)


eslesar:
Agreed. Thanks

PR #454, opened 2016-06-29
Wmf restructure

juanpablojofre:
New structure to allow for new versions

---
This change is [Reviewable](https://reviewable.io/reviews/powershell/powershell-docs/454)


eslesar:
This needs to be backed out. It looks like the restructuring was done without getting the most recent content--so a bunch of changes I did last week to the dsc wmf topics would be clobbered by this.

PR #493, opened 2016-07-06
fixed broken links

eslesar:
This change is [Reviewable](https://reviewable.io/reviews/powershell/powershell-docs/493)


vors:
Doesn't OPS populate relative links automatically?

eslesar:
Only if the topics are within the same docset. Between docsets, we have to use the URL of the published topic. Because these are between the wmf and dsc docsets, they need to be URLs.

vors:
I see, thank you for the explanation @eslesar !
Could we potentially teach OPS understands links in the same repo?

The reason is: if OPS can do that, then we get parity between navigating markdown sources on github and navigating published version from OPS.

joeyaiello:
@vors: I believe we have an eventual plan to have the entire repo correspond to a single docset rather than multiple to solve problems like this.

PR #529, opened 2016-07-20
Removing duplicate files and renaming the correct one

HemantMahawar:
This change is [Reviewable](https://reviewable.io/reviews/powershell/powershell-docs/529)


HemantMahawar:
@jaimeo @BerheAbrha @Indhukrishna  ... Does this change looks good?

Issue #503, opened 2016-07-11
OPS doesn't render pipe symbol correctly


HemantMahawar:
Details in #501

joeyaiello:
Fixed by #536

PR #501, opened 2016-07-11
Replaced pipe character

JeffWouters:
Using the pipe messes up the table view for that row.

---
This change is [Reviewable](https://reviewable.io/reviews/powershell/powershell-docs/501)

PR #536, opened 2016-07-26
Change in markdown formatting of the table

KirkMunro:
The table used a pipe symbol within a code block. When this document is used to generate content for MSDN, that causes the table to render incorrectly. Using <code></code> to identify the code block instead of backticks and replacing the internal pipe with &#124; (ascii pipe) should resolve this issue.

---
This change is [Reviewable](https://reviewable.io/reviews/powershell/powershell-docs/536)


zjalexander:
for what it's worth, based on #503 and #501, I'm not sure this is guaranteed to fix the issue.

PR #566, opened 2016-08-04
Change topic title to be appropriate for MSDN

JKeithB:
Current content came from welcome page of Gallery, so starts with "Welcome to...", which is inappropriate for MSDN docs for the Gallery.

---
This change is [Reviewable](https://reviewable.io/reviews/powershell/powershell-docs/566)


AsCalvo:
Keith - can you please help me understand the reason for reversing the changes? And why we're using content from MSDN instead of having unique GitHub content? Thanks!

PR #600, opened 2016-08-23
Fix backslashes

juanpablojofre:
This patch should address all wrongly added backslashes during the content port from Technet

---
This change is [Reviewable](https://reviewable.io/reviews/powershell/powershell-docs/600)


juanpablojofre:
@eslesar , please take a look.
This should address some of the long standing open issues.

PR #622, opened 2016-09-15
Updated examples, added fix to metaconfig

eslesar:


eslesar:
@jpjofre please review

PR #670, opened 2016-10-07
Documenting DSCAutomationHostEnabled registry key

nitinbps:
In response to http://answers.microsoft.com/en-us/windows/forum/windows_10-other_settings/some-settings-are-managed-by-your-organization-on/7d0e0fa3-50e2-4a56-8e91-8539b5fbe648?auth=1.

### DIFF
--- a/dsc/DSCAutomationHostEnabled.md
+++ b/dsc/DSCAutomationHostEnabled.md
@@ -1,0 +1,11 @@
+
+DSC uses <b>DSCAutomationHostEnabled </b>registry key under <b>HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Policies\System</b> to automatically configure the machine at initial boot-up.
+DSCAutomationHostEnabled supports three modes:
+
+|  DSCAutomationHostEnabled Value  |  Description   | 
+|---|---| 
+0 | Disable configuring the machine at boot-up. |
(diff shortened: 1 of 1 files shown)


ArieHein:
- Should probably change the syntax , swapping DSC to LCM :
+2 | Enable configuring the machine only if LCM is in pending or current state. This is the default value.|

- Should probably change the default value to 0 on a Windows Home Edition. As it may lead to some security issues. Or in fact on any windows including servers so it will be used because I "Wanted" and not because it was already active.

- If I set it to 0 , how does this affect 3rd party configuration managers ? This should override their clients as well I imagine ?

- If I set it to 0 and the LCM is set to Pull, then at boot time, the LCM will not go to the pull server to seek new configurations. That's the purpose ? If so, its a nicer way to disable LCM then sending a new configuration to the node :)  (hope this doesn't lead to missuse or malicious things)

- If I set it to 0, will the LCM still do the state check on itself (as its a scheduled task) ?
 Will it send events to the pull server (just not apply the configuration)? or is it completely dead? 

- What other states can LCM be in, that I would want the configuration to start anyway ? (basically the difference in real environments between 1 and 2)

Somehow I sense a blog post on the PowerShell team blog about this, with your answers ;)

Thanks

nitinbps:
Yes, this entry itself won't suffice. We have to write a blog explaining the feature itself first.

PR #689, opened 2016-10-10
new docset: reference

saldana:


juanpablojofre:
Sandra,
We usually don't push requests into Live from a branch that is not Staging.
And, the request to merge Reference is now failing; due to errors in Reference that we didn't see before as it wasn't a docset.
Once Reference gets fixed and passes all checks, then we'll do the merge to Live.
Until then, I'll close this request.

PR #696, opened 2016-10-11
Update link to WMF

joeyaiello:
Unfortunately, I couldn't find a fwdlink that pointed to WMF 5.0, but hopefully this is still better than a broken link.


HemantMahawar:
@joeyaiello @zjalexander I created a FW link to WMF 5.0 DLC page and updated the doc.

PR #722, opened 2016-10-18
Removing the feature-scenario-cache folder

eslesar:


eslesar:
@juanpablojofre please review

PR #724, opened 2016-10-22
Update README.md

HemantMahawar:
With WMF 5.0 url


HemantMahawar:
Will submit again, later

PR #708, opened 2016-10-13
Emphasizing how to acquire PS 5

zjalexander:
SkyEye feedback shows user issues finding the download link

PR #725, opened 2016-10-25
Update Windows-PowerShell-System-Requirements.md

Jagoop:
Added PowerShell 5.0 requirements

### DIFF
--- a/scripting/setup/Windows-PowerShell-System-Requirements.md
+++ b/scripting/setup/Windows-PowerShell-System-Requirements.md
@@ -16,6 +16,20 @@ This topic lists the system requirements for Windows PowerShell 3.0 and Windows 
 Windows® 8.1 and Windows Server® 2012 R2 include all required programs. This topic is designed for users of earlier releases of Windows.
 
 ## Operating System Requirements
+Windows PowerShell 5.0 runs on the following versions of Windows.
+
+-   Windows Server 2016, installed by default
+
+-   Windows Server 2012 R2
+
+-   Windows Server 2012
+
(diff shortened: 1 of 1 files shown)


scripting/setup/Windows-PowerShell-System-Requirements.md
```diff
@@ -16,6 +16,20 @@ This topic lists the system requirements for Windows PowerShell 3.0 and Windows
...
+
+-   Windows Server 2012 R2
+
+-   Windows Server 2012
+
+-   Windows Server 2008 R2 with Service Pack 1, install [Windows Management Framework 4.5](http://go.microsoft.com/fwlink/?LinkID=242919) to run Windows PowerShell 4.0
+
+-   Windows 8.1
+
+-   Windows 7 with Service Pack 1, install [Windows Management Framework 4.5](http://go.microsoft.com/fwlink/?LinkID=242919) to run Windows PowerShell 4.0
```
kilasuit: Same as above

kilasuit:
@zjalexander - this seems to be an almost defunct file due to the issues I raised in #708 - as the requirements page https://msdn.microsoft.com/en-us/powershell/wmf/5.0/requirements isn't being pointed to

Can we please get this clarified with priority as there is little point having an additional page of documentation for sake of having it if it correctly documented on another page else where.

This is causing unneeded confusion in a number of communities because 1 page incorrectly says 1 thing whilst another says something different (but actually the right info)

zjalexander:
cc @juanpablojofre 
personally i think there's value in having the 5.0 requirements page since it does get linked directly from ... somewhere (need to do some digging to remember where).
we should have both "general system requirements" (edited in this PR) as well as the 5.0 page, and both should be accurate.

PR #738, opened 2016-10-26
Updating Live with reference content

juanpablojofre:


juanpablojofre:
@eslesar , @HemantMahawar , 
Gentlemen, what are we waiting here?
Would any of you merge this PR or state why we shouldn't merge it.

PR #743, opened 2016-10-28
Register-AugumentCompleter per work item 69310.

TimShererAtAquent:
Update Register-AugumentCompleter per work item 69310. 

Changes to example section and input and output sections.


juanpablojofre:
@TimShererAtAquent , thanks a lot!!

Issue #751, opened 2016-11-01
Missing files are referenced in 5.1/ToC.md 


HemantMahawar:
The reference\5.1\Toc.md refers to following files
Microsoft.powershell.security/Providers/Get-Childitem-for-Certificate.md
Microsoft.powershell.security/Providers/Move-item-for-Certificate.md
Microsoft.powershell.security/Providers/New-Item-for-Certificate.md
Microsoft.powershell.security/Providers/Remove-Item-for-Certificate.md

but they are missing in from [reference folder](https://github.com/PowerShell/PowerShell-Docs/tree/staging/reference/5.1/Microsoft.PowerShell.Security/Providers), instead it has wrong files (may be from [WsMan provider](https://github.com/PowerShell/PowerShell-Docs/tree/staging/reference/5.1/Microsoft.WsMan.Management/Providers)) folder

@juanpablojofre

HemantMahawar:
@juanpablojofre  #737 makes it more wrong :)

In 5.1 TOC, under Microsoft.PowerShell.Security --> Providers you now have

Certificate-Provider
New-Item-for-ClientCertificate
New-Item-for-InitializationParameters
New-Item-for-Listener
New-Item-for-Plugin
New-Item-for-Resources
New-Item-for-Security

but the right thing is to have it similar to 5.0 TOC, under Microsoft.PowerShell.Security --> Providers should have

Certificate Provider
Get-Childitem-for-Cerificate
Move-item-for-Cerificate
New-item-for-Cerificate
Remove-item-for-Cerificate

juanpablojofre:
@HemantMahawar ,
It all fixed.
Could you please review and close this issue, if your happy with the fix.
If not, leave your comments; so, I can improve the fix.

HemantMahawar:
I don't see any more changes related to this after #757. See my above comment on why #757 doesn't fix it

juanpablojofre:
@HemantMahawar ,
Can you tell us if this issue is now resolved?
Thanks,
-- JP

HemantMahawar:
@juanpablojofre  Yup, its fixed now. Thanks for the quick action here :)

HemantMahawar:
Fixed via #761

PR #760, opened 2016-11-03
Moving latest content to Live

juanpablojofre:


juanpablojofre:
Waiting on #751 to be fixed

PR #1851, opened 2017-11-11
Fix the example of `Format-Wide -ShowError`

matt9ucci:
The example throws exception because its `-Property` parameter accepts an Object, not an array:

```powershell
PS> Get-Date | Format-Wide DayOfWeek,{ $_ / $null } -ShowError
Format-Wide : Cannot convert System.Object[] to one of the following types {System.String, System.Management.Automation.ScriptBlock}.
At line:1 char:12
+ Get-Date | Format-Wide DayOfWeek,{ $_ / $null } -ShowError
+            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidArgument: (:) [Format-Wide], NotSupportedException
    + FullyQualifiedErrorId : DictionaryKeyUnknownType,Microsoft.PowerShell.Commands.FormatWideCommand
```

- [x] Impacts 6 document
- [x] Impacts 5.1 document
- [x] Impacts 5.0 document
- [x] Impacts 4.0 document
- [x] Impacts 3.0 document

### DIFF
--- a/reference/3.0/Microsoft.PowerShell.Utility/Format-Wide.md
+++ b/reference/3.0/Microsoft.PowerShell.Utility/Format-Wide.md
@@ -208,18 +208,18 @@ Accept wildcard characters: True
 ```
 
 ### -ShowError
-Sends errors through the pipeline.
-This parameter is rarely used, but can be used as a debugging aid when you are formatting expressions in a Format-Wide command, and the expressions do not appear to be working.
-The following shows an example of the results of adding the ShowError parameter with an expression.
+Indicates that the cmdlet sends errors through the pipeline.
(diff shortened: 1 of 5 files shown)


sdwheeler:
@matt9ucci Thanks for the submission. This is another one that should use `-DisplayError` instead of `-ShowError`. Can you make that fix as well?

matt9ucci:
@sdwheeler Although it might be somewhat confusing, this example is for `-ShowError` parameter, not for `-DisplayError`.

sdwheeler:
@matt9ucci - Oh, got it. I didn't notice at first that this cmdlet has both parameters.

Issue #2294, opened 2018-04-03
CIM cmdlets help topics are only available for version 6, creating the mistaken impression that they were only introduced then


mklement0:
CIM cmdlets were introduced in v3, yet this repo only contains v6 topics.

This also affects the published topics in that they mistakenly suggest that CIM cmdlets were only introduced in v6:

E.g., attempting to switch to the v3 topic for `Get-CimClass` redirects to https://docs.microsoft.com/en-us/powershell/module/cimcmdlets/get-cimclass?view=powershell-6&viewFallbackFrom=powershell-3.0 which then displays the following, misleading message:

> The requested page is not available for Powershell 3.0. You have been redirected to the newest product version this page is available for.

Version(s) of document impacted
------------------------------
- [ ] Impacts 6.1 document
- [ ] Impacts 6.0 document
- [x] Impacts 5.1 document
- [x] Impacts 5.0 document
- [x] Impacts 4.0 document
- [x] Impacts 3.0 document

Reason(s) for not selecting all version of documents
----------------------------------------------------

The v6 topics are present, but no other versions.

sdwheeler:
Need to investigate if the docs existed and were delete or if the docs were never published. If deleted, then we need to try to restore. If never existed then we need to recreate the old versions.

sdwheeler:
Added CimCmdlets and PSDiagnostics to all versions.

PR #2383, opened 2018-05-01
Adding missing docs to version 3 content

sdwheeler:
Added help files for:
- CimCmdlets
- PSDiagnostics

These were missing from the docs but shipped in v3.

See issue #2294

- [ ] Impacts 6.1 document
- [ ] Impacts 6.0 document
- [ ] Impacts 5.1 document
- [ ] Impacts 5.0 document
- [ ] Impacts 4.0 document
- [x] Impacts 3.0 document

- [x] This PR partially fixes the issue, and issue #2294 tracks the remaining work


zjalexander:
A _lot_ of the files have placeholder content. Is that in the case in later versions as well? If so, it would be great if we could backport that in. If not, we should really try and get the content complete before we integrate it.

sdwheeler:
@zjalexander The placeholder content is all in the psdiagnostics module cmdlets. It is the same for every version. So integrating it now would be no worse than the version published for v6. But I can remove it if you think that we should wait.

sdwheeler:
@zjalexander Per our discussion, I am merging this and have created issue #2404.

PR #2575, opened 2018-06-14
 Fixes #2416 update about_comparison_operators to reflect regular expression substitution

### LINKED ISSUE #2416: Explain the features and syntax of the -replace operator's replacement operand
PowerShell's `-replace` operator is based on the .NET `[regex]::Replace()` method, and the syntax of the replacement operand is currently not explained or linked to in [`about_Comparison_Operators`](https://github.com/PowerShell/PowerShell-Docs/blob/staging/reference/6/Microsoft.PowerShell.Core/About/about_Comparison_Operators.md)

Not only may users therefore miss out on referencing capture groups (e.g, `$1`) in the replacement operand, they may mistakenly assume that it is a _literal_ and be surprised when `$` chars. are subject to interpretation, notably a `$$` turning into a _single_ `$` (because _doubling_ `$` is what it takes to escape them).

```powershell
PS> 'foo' -replace 'foo', 'f$$'
f$   # '$$' was interpreted as an *escaped* $
```

To use an arbitrary value (supplied via variable whose value isn't known ahead of time) as the replacement operand, use the following:

```powershell
PS> $var = 'f$$'; 'foo' -replace 'foo', $var.Replace('$', '$$')
f$$
```

`about_Comparison_Operators` should be updated to:

* at least briefly mention the syntax and capabilities of the replacement operand
* link to the .NET documentation at https://docs.microsoft.com/en-us/dotnet/standard/base-types/substitutions-in-regular-expressions
* show the above technique for treating a replacement operand as a _literal_.

- [x] Impacts 6.1 document
- [x] Impacts 6.0 document
- [x] Impacts 5.1 document
- [x] Impacts 5.0 document
- [x] Impacts 4.0 document
- [x] Impacts 3.0 document

bobbytreed:
- [ ] Impacts 6.1 document
- [X] Impacts 6.0 document
- [X] Impacts 5.1 document
- [X] Impacts 5.0 document
- [X] Impacts 4.0 document
- [X] Impacts 3.0 document


sdwheeler:
Clean up your branch first.

PR #2576, opened 2018-06-14
 Fixes #694 update about_requires to reflect current functionality

### LINKED ISSUE #694: #Requires -Module runs first, even when it's not first line in a script (not doc'd)
In about_Requires, we need to add a line explaining that this directive runs first in the script, even when it doesn't appear on the first line in the script.

`#Requires` is supposed to be the first line in the script (as doc'd). But, if it's not, there's no error. Instead, it runs first without warning. Therefore, code like this, in which order matters, fails unexpectedly, because the correct version of the module is imported, then removed.

```powershell
Get-Module Pester | Remove-Module   #Removes current version
#Requires -Module @{ModuleName = 'Pester'; ModuleVersion = '3.4.3'}
...
```

bobbytreed:
- [ ] Impacts 6.1 document
- [ ] Impacts 6.0 document
- [X] Impacts 5.1 document
- [X] Impacts 5.0 document
- [ ] Impacts 4.0 document
- [X] Impacts 3.0 document


sdwheeler:
Clean up your branch first.

PR #2674, opened 2018-07-08
Remove duplicate sentences in developer\cmdlet\adding-*.md

matt9ucci:
These duplicates might be due to failure of auto-conversion of msdn links.


DCtheGeek:
@matt9ucci And thanks for another set of contributions!  Really appreciate the help cleaning up this content.

PR #3269, opened 2018-11-12
Added the MyTypeName special variable.

randomnote1:
Added information about the `$MyTypeName` special variable which is only available within a DSC configuration. The initial research was conducted and published here: https://blogs.msdn.microsoft.com/powersql/2017/10/31/dynamically-determine-a-dsc-configuration-name-inside-a-configuration/


sdwheeler:
@bobbytreed Please review

Issue #3740, opened 2019-02-20
Placeholder tags need to be filled in with good content


bobbytreed:
reference\5.0\PackageManagement\Find-Package.md
reference\5.0\PackageManagement\Get-Package.md
reference\5.0\PackageManagement\Get-PackageSource.md
reference\5.0\PackageManagement\Install-Package.md
reference\5.0\PackageManagement\Register-PackageSource.md
reference\5.0\PackageManagement\Save-Package.md
reference\5.0\PackageManagement\Set-PackageSource.md
reference\5.0\PackageManagement\Uninstall-Package.md
reference\5.0\PackageManagement\Unregister-PackageSource.md
reference\5.0\PowershellGet\Save-Module.md
reference\5.0\PowershellGet\Uninstall-Module.md
reference\5.1\PackageManagement\Find-Package.md
reference\5.1\PackageManagement\Get-Package.md
reference\5.1\PackageManagement\Get-PackageSource.md
reference\5.1\PackageManagement\Install-Package.md
reference\5.1\PackageManagement\Register-PackageSource.md
reference\5.1\PackageManagement\Save-Package.md
reference\5.1\PackageManagement\Set-PackageSource.md
reference\5.1\PackageManagement\Uninstall-Package.md
reference\5.1\PackageManagement\Unregister-PackageSource.md
reference\5.1\PowershellGet\Find-Command.md
reference\5.1\PowershellGet\Find-DscResource.md
reference\5.1\PowershellGet\Find-RoleCapability.md
reference\5.1\PowershellGet\New-ScriptFileInfo.md
reference\5.1\PowershellGet\Publish-Module.md
reference\5.1\PowershellGet\Save-Module.md
reference\5.1\PowershellGet\Save-Script.md
reference\5.1\PowershellGet\Uninstall-Module.md
reference\5.1\PowershellGet\Uninstall-Script.md
reference\5.1\PowershellGet\Update-Module.md
reference\5.1\PowershellGet\Update-ModuleManifest.md
reference\5.1\PowershellGet\Update-Script.md
reference\5.1\PowershellGet\Update-ScriptFileInfo.md
reference\6\PackageManagement\Find-Package.md
reference\6\PackageManagement\Get-Package.md
reference\6\PackageManagement\Get-PackageSource.md
reference\6\PackageManagement\Install-Package.md
reference\6\PackageManagement\Register-PackageSource.md
reference\6\PackageManagement\Save-Package.md
[...]

PR #4517, opened 2019-07-03
Fixed empty parameters, copyedits

davidsmatlak:
- [x] Impacts 7 document
- [x] Impacts 6 document
- [x] Impacts 5.1 document
- [x] Impacts 5.0 document
- [ ] Impacts 4.0 document
- [ ] Impacts 3.0 document

Acrolinx scores
Save-Script: original version: 95, updated version: 98
Uninstall-Module: original version: 93, updated version: 97
Uninstall-Script: original version: 94, updated version: 97

- Fixes [AB#1565976](https://dev.azure.com/mseng/677da0fb-b067-4f77-b89b-f32c12bb8617/_workitems/edit/1565976)
- Related to #3740
- Added content to empty parameter descriptions
- Copyedits, style
- Revised the Save-Module parameter descriptions: MaximumVersion, MinimumVersion, and Path

### DIFF
--- a/reference/5.0/PowershellGet/Uninstall-Module.md
+++ b/reference/5.0/PowershellGet/Uninstall-Module.md
@@ -1,5 +1,5 @@
 ---
-ms.date: 7/2/2019
+ms.date: 07/02/2019
 schema:  2.0.0
 locale:  en-us
 keywords:  powershell,cmdlet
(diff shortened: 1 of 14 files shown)


reference/5.0/PowershellGet/Uninstall-Module.md
```diff
@@ -16,33 +16,49 @@ Uninstalls a module.
...
-The **Uninstall-Module** cmdlet uninstalls the specified module from the local computer.
-You cannot uninstall a module if it has other modules as dependencies.
+
+The `Uninstall-Module` cmdlet uninstalls a specified module from the local computer. You can't
+uninstall a module if it has other modules as dependencies.
 
 ## EXAMPLES
 
-### Example 1: Get a module and uninstall it
+### Example 1: Uninstall a module
```
sdwheeler: Why did you want to change the example to not use the pipeline? It is a good use case. Also, we don't need to call Get-InstalledModule if we are not using the pipeline.

I think we need one example that is just a simple uninstall module by name. and one example using the pipeline.

reference/5.1/PowershellGet/Uninstall-Script.md
```diff
@@ -12,36 +12,52 @@ title: Uninstall-Script
...
+The `Uninstall-Script` cmdlet uninstalls a specified script from the local computer.
 
 ## EXAMPLES
 
 ### Example 1: Uninstall a script file
+
+This example uninstalls a script from the local computer.
+
+```powershell
+Get-InstalledScript -Name UpdateManagement-Template
```
sdwheeler: We don't need to call Get-InstalledScript if we are not using the pipeline. Think we should have a simple example to show uninstall by name and another to show using the pipeline.

reference/5.0/PowershellGet/Uninstall-Module.md
```diff
@@ -1,5 +1,5 @@
 ---
-ms.date:  06/09/2017
+ms.date: 7/2/2019
```
sdwheeler: Date needs to be 07/02/2019. Need all the digits. Fix in all articles.

reference/5.0/PowershellGet/Uninstall-Module.md
davidsmatlak: My initial tests were incorrect because the pipeline didn't work, so I split the commands. I read your feedback and retested, and the pipeline tests were successful. I updated the examples.

reference/5.1/PowershellGet/Uninstall-Script.md
davidsmatlak: Thanks for the feedback. I updated the examples.

reference/5.0/PowershellGet/Uninstall-Module.md
davidsmatlak: Thanks. I updated the date in each document.

PR #4553, opened 2019-07-11
Add link to online version

skycommand:
- [X] Impacts 7 document
- [X] Impacts 6 document
- [X] Impacts 5.1 document
- [X] Impacts 5.0 document
- [X] Impacts 4.0 document
- [X] Impacts 3.0 document

### DIFF
--- a/reference/5.1/Microsoft.PowerShell.Core/About/about_Remote_Troubleshooting.md
+++ b/reference/5.1/Microsoft.PowerShell.Core/About/about_Remote_Troubleshooting.md
@@ -3,6 +3,7 @@ ms.date:  01/18/2019
 schema:  2.0.0
 locale:  en-us
 keywords:  powershell,cmdlet
+online version: https://docs.microsoft.com/powershell/module/microsoft.powershell.core/about/about_remote_troubleshooting
 title:  about_Remote_Troubleshooting
 ---
 # About Remote Troubleshooting


sdwheeler:
Thank you for the contribution. Unfortunately I can't accept the change for a couple of reasons.
1. The `online version` metadata value must be an FWLink URL. Those can only be created by us internally.
1. The `online version` metadata value does not apply to About topic because those help assets are published as text files and not the XML format of cmdlet help.

skycommand:
@sdwheeler 
> 1. The `online version` metadata value must be an FWLink URL. Those can only be created by us internally.

Must? Well, currently, this page has the following link associated with it:

    http://technet.microsoft.com/library/hh847850(v=wps.630).aspx

Here is a screenshot:

Screenshot

I filed this request because that link is wrong.

> 2. The `online version` metadata value does not apply to About topic because those help assets are published as text files and not the XML format of cmdlet help.

So, its current metadata like `ms.date`, `schema` and `locale` are worthless? I did not know that. Still, since this GitHub repo is the official feedback channel for the documentation, here is mine: Please made `about_` pages accessible via the `Get-Help -online` cmdlet. It helps us, administrators, a lot. Thanks.

sdwheeler:
@skycommand Yes, I know there are a lot of broken links. FWLinks are the current standard we are using for `online verison`. The tool to create and manage FWLink is internal to Microsoft only. I could create one and update the link in the document. However, our build pipeline is not publishing updated help CAB files for any version of PowerShell except v6. So, even if I accepted your PR, you would not benefit from the change.

This a known infrastructure issue that we want to fix. It is an issue of time, priorities, and internal dev resources to fix the pipeline. Believe me, I wish I could fix this.

skycommand:
Ouch! So, I must probably stop issuing Update-Help on a monthly basis.

PR #4575, opened 2019-07-17
import-dscresource on separate lines

mgreenegit:
- [ ] Impacts 7 document
- [ ] Impacts 6 document
- [ ] Impacts 5.1 document
- [ ] Impacts 5.0 document
- [ ] Impacts 4.0 document
- [ ] Impacts 3.0 document


dsc/configurations/write-compile-apply-configuration.md
```diff
@@ -42,6 +42,11 @@ Configuration HelloWorld {
 }
 ```
 
+> !Important
+> In more advanced scenarios where multiple modules need to be imported so you can work with many DSC Resources
+> in the same configuration, make sure to put each module in a seperate line using `Import-DscResource`.
```
bobbytreed: Could you add a code block here, showing the difference or an example of what you mean?

mgreenegit: trying code block within note.  if that doesn't work I will take it out of note annotation.

PR #4655, opened 2019-08-19
Update pullServer.md

darkocobe:
I guess in the example was just typo showing the correct version numbers. "For example, a module named xWebAdminstration with a module version of 3.1.2.0 would be named `xWebAdministration_3.2.1.0.zip`." If this is the case should be fixed otherwise will be confusing for people naming the modules. They will think that they need to swap the numbers.

- [ ] Impacts 7 document
- [ ] Impacts 6 document
- [ ] Impacts 5.1 document
- [ ] Impacts 5.0 document
- [ ] Impacts 4.0 document
- [ ] Impacts 3.0 document


sdwheeler:
Triggering new build

PR #4723, opened 2019-09-02
Change casing in example

DanAtkinson:
Change instances of `get-process` to `Get-Process`.

- [ ] Impacts 7 document
- [ ] Impacts 6 document
- [ ] Impacts 5.1 document
- [ ] Impacts 5.0 document
- [ ] Impacts 4.0 document
- [ ] Impacts 3.0 document

### DIFF
--- a/reference/6/Microsoft.PowerShell.Core/About/about_Hash_Tables.md
+++ b/reference/6/Microsoft.PowerShell.Core/About/about_Hash_Tables.md
@@ -294,8 +294,8 @@ The following statement creates a hash table of process name strings and
 process object values and saves it in the \$p variable.
 
 ```powershell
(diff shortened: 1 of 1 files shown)


DanAtkinson:
@sdwheeler I have absolutely no idea what the warnings from @opbld33 relate to, and the build report isn't very helpful. It says that locale is not allowed to be set but it's set at the header of every document.

sdwheeler:
You can ignore those warnings.

The build automation that is reporting the warning because it doesn't want the locale in the header. However, it is needed for other purposes downstream. I have filed a bug to have the build automation ignore that warning. 

If there are errors or warnings that need your attention when you submit a PR, I will let you know and ask you to make a change.

PR #4777, opened 2019-09-11
Fixed -ToSession description

Torch02:
The -ToSession description seems to have been a copy/paste from the -FromSession description. When using the -ToSession parameter, the -Destination parameter becomes local to the remote computer. This updates fixes that copy/paste error & provides the correct information.

- [ ] Impacts 7 document
- [ ] Impacts 6 document
- [ ] Impacts 5.1 document
- [ ] Impacts 5.0 document
- [ ] Impacts 4.0 document
- [ ] Impacts 3.0 document


sdwheeler:
Actually, my understanding is that when you use `-ToSession`, **Path** and **LiteralPath** parameters refer to the local path on the remote machine in that session.

PR #4795, opened 2019-09-16
Update build to use Pandoc 2.7.3

sdwheeler:
Updated build to use Pandoc 2.7.3

Added tools/build-updatedhelp.ps1 - This script can be used to build the updateable help for a single version folder. Using this script, a user could build and install any version of help they need.

Usage:

1. Clone the docs respository

   ```
    cd C:\temp
    git clone https://github.com/MicrosoftDocs/PowerShell-Docs.git
   ```

2. Run build-updatedhelp.ps1 for the version folder you want to build

   ```
    .\build-updatedhelp.ps1 -sourceFolder C:\temp\PowerShell-Docs\reference\5.1 -Verbose
   ```

3. Run Update-Help  to install the newly built help

   ```
    Update-Help -SourcePath c:\temp\updatablehelp\5.1 -Recurse -Force
   ```

### DIFF
--- a/tools/build-updatedhelp.ps1
+++ b/tools/build-updatedhelp.ps1
@@ -21,6 +21,10 @@ param(
     [string]$sourceFolder
 )
 
+# Turning off the progress display, by default
+$savedProgressPreference = $global:ProgressPreference
+$global:ProgressPreference = 'SilentlyContinue'
+
 $tempDir = [System.IO.Path]::GetTempPath()
 
 # Pandoc source URL
@@ -139,3 +143,5 @@ Get-ChildItem $VersionFolder -Directory | ForEach-Object -Process {
         Write-Error -Message "PlatyPS failure: $ModuleName -- $Version" -Exception $_
     }
 }
+
+$global:ProgressPreference = $savedProgressPreference
--- a/tools/build-updatedhelp.ps1
+++ b/tools/build-updatedhelp.ps1
@@ -5,9 +5,9 @@
        cd C:\temp
        git clone https://github.com/MicrosoftDocs/PowerShell-Docs.git
 
-    2. Run make-updatedhelp.ps1 for the version folder you want to build
+    2. Run build-updatedhelp.ps1 for the version folder you want to build
 
-      .\make-updatedhelp.ps1 -sourceFolder C:\temp\PowerShell-Docs\reference\5.1 -Verbose
+      .\build-updatedhelp.ps1 -sourceFolder C:\temp\PowerShell-Docs\reference\5.1 -Verbose
 
     3. Run Update-Help  to install the newly built help
 
@@ -68,7 +68,7 @@ function Get-ContentWithoutHeader {
         }
     }
     if ($end -gt $start) {
-        Write-Output ($doc[$end..$($doc.count)] -join "`r`n")
+        Write-Output ($doc[$end..$($doc.count)] -join ([Environment]::Newline))
     }
     else {
         Write-Output ($doc -join "`r`n")
--- a/tools/make-updatedhelp.ps1
+++ b/tools/make-updatedhelp.ps1
@@ -1,0 +1,141 @@
+<#
+    Usage:
+
+    1. Clone the docs respository
+       cd C:\temp
+       git clone https://github.com/MicrosoftDocs/PowerShell-Docs.git
+
+    2. Run make-updatedhelp.ps1 for the version folder you want to build
+
+      .\make-updatedhelp.ps1 -sourceFolder C:\temp\PowerShell-Docs\reference\5.1 -Verbose
+
+    3. Run Update-Help  to install the newly built help
+
+       Update-Help -SourcePath c:\temp\updatablehelp\5.1 -Recurse -Force
+#>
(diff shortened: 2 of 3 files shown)


tools/make-updatedhelp.ps1
```diff
@@ -0,0 +1,141 @@
+<#
+    Usage:
+
+    1. Clone the docs respository
+       cd C:\temp
+       git clone https://github.com/MicrosoftDocs/PowerShell-Docs.git
+
+    2. Run make-updatedhelp.ps1 for the version folder you want to build
+
+      .\make-updatedhelp.ps1 -sourceFolder C:\temp\PowerShell-Docs\reference\5.1 -Verbose
```
SteveL-MSFT: `make` isn't a valid verb, perhaps `build`?

tools/make-updatedhelp.ps1
```diff
@@ -0,0 +1,141 @@
...
+            else {
+                if ($end -eq -1) {
+                    $end = $x + 1
+                    break
+                }
+            }
+        }
+    }
+    if ($end -gt $start) {
+        Write-Output ($doc[$end..$($doc.count)] -join "`r`n")
```
SteveL-MSFT: Would be better to use `[Environment]::Newline` unless the tooling only works with Windows style newlines

tools/make-updatedhelp.ps1
```diff
@@ -0,0 +1,141 @@
...
+        # For each module, create update-help help files (cab and helpinfo.xml files)
+        $cabInfo = New-ExternalHelpCab -CabFilesFolder $MamlOutputFolder -LandingPagePath $LandingPage -OutputFolder $CabOutputFolder
+
+        # Only output the cab fileinfo object
+        if ($cabInfo.Count -eq 8) { $cabInfo[-1].FullName }
+    }
+    catch {
+        Write-Error -Message "PlatyPS failure: $ModuleName -- $Version" -Exception $_
+    }
+}
```
SteveL-MSFT: Best practice is to have a trailing newline

tools/make-updatedhelp.ps1
```diff
@@ -0,0 +1,141 @@
...
+    try {
+        # For each module, create a single maml help file
+        # Adding warningaction=stop to throw errors for all warnings, erroraction=stop to make them terminating errors
+        New-ExternalHelp -Path $ModulePath -OutputPath $MamlOutputFolder -Force -WarningAction Stop -ErrorAction Stop
+
+        # For each module, create update-help help files (cab and helpinfo.xml files)
+        $cabInfo = New-ExternalHelpCab -CabFilesFolder $MamlOutputFolder -LandingPagePath $LandingPage -OutputFolder $CabOutputFolder
+
+        # Only output the cab fileinfo object
+        if ($cabInfo.Count -eq 8) { $cabInfo[-1].FullName }
```
SteveL-MSFT: Where does the magic number 8 come from?

tools/make-updatedhelp.ps1
sdwheeler: renamed

tools/make-updatedhelp.ps1
sdwheeler: The intent of this tool is Windows-only but I will change it. It could be targeted at any docs for any platform.

tools/make-updatedhelp.ps1
sdwheeler: Where? At the end of the script file? I added a blank line to the file. Let me know if you meant something else.

tools/make-updatedhelp.ps1
sdwheeler: This is just copied from the existing build script. It how `New-ExternalHelpCab` returns its data. I assume we are checking that it returned the expected amount of data so we can display the interesting piece.

tools/build-updatedhelp.ps1
```diff
@@ -0,0 +1,141 @@
...
+$tempDir = [System.IO.Path]::GetTempPath()
+
+# Pandoc source URL
+$panDocVersion = "2.7.3"
+$pandocSourceURL = "https://github.com/jgm/pandoc/releases/download/$panDocVersion/pandoc-$panDocVersion-windows-x86_64.zip"
+
+$docToolsPath = New-Item (Join-Path $tempDir "doctools") -ItemType Directory -Force
+$pandocZipPath = Join-Path $docToolsPath "pandoc-$panDocVersion-windows-x86_64.zip"
+Write-Verbose "Downloading Pandoc..."
+Invoke-WebRequest -Uri $pandocSourceURL -OutFile $pandocZipPath
```
SteveL-MSFT: Not a huge deal, but on WinPS5.1 progress slows down invoke-webrequest and expand-archive quite a bit.  This isn't an issue in PSCore6.  Don't know how big the zip is so it may not be a big issue even on 5.1.  Otherwise, you may want to special case 5.1 by changing progresspreference to silentlycontinue and then changing it back to the original value.

sdwheeler: Fixed

PR #4856, opened 2019-09-30
Document pipeline chain operators (&&/||)

rjmholt:
Resolves https://github.com/MicrosoftDocs/PowerShell-Docs/issues/4853.

Document pipeline chain operators (`&&`/`||`), which are added to PowerShell 7 by https://github.com/PowerShell/PowerShell/pull/9849.

See also: https://github.com/PowerShell/PowerShell-RFC/pull/192.

- [x] Impacts 7 document
- [ ] Impacts 6 document
- [ ] Impacts 5.1 document
- [ ] Impacts 5.0 document
- [ ] Impacts 4.0 document
- [ ] Impacts 3.0 document

- [x] The documented feature was introduced in version (list version here) of PowerShell


reference/7/Microsoft.PowerShell.Core/About/about_Pipeline_Chain_Operators.md
```diff
@@ -14,27 +14,32 @@ Describes chaining pipelines with the `&&` and `||` operators in PowerShell.
...
+[AND-OR lists](https://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html#tag_02_09_03)
+in POSIX shells like bash, zsh and sh, as well as
+[conditional processing symbols](/previous-versions/windows/it-pro/windows-xp/bb490954(v=technet.10)#using-multiple-commands-and-conditional-processing-symbols)
+in the Windows Command Shell (cmd.exe).
+
+The `&&` operator executes the right-hand pipeline, if the left-hand pipeline
+succeeded. Conversely, the `||` operator executes the right-hand pipeline if
+the left-hand pipeline failed.
+
+This can be use for conditional execution of software deployments. For example:
```
rjmholt: `use` -> `used`.

The language here feels too specific for an operator to me.

If this example demands language like that, we should change the example.

I want to be able to emphasise that this works with native commands based on `$?`/`$LASTEXITCODE` and not just with cmdlets/functions.

reference/7/Microsoft.PowerShell.Core/About/about_Pipeline_Chain_Operators.md
```diff
@@ -14,27 +14,32 @@ Describes chaining pipelines with the `&&` and `||` operators in PowerShell.
...
 ```
 
-The `&&` operator will execute the right-hand pipeline,
-if the left-hand pipeline succeeded,
-whereas the `||` operator will execute the right-hand pipeline
-if the left-hand pipeline failed.
+In this example the `npm run deploy` command only runs if the build command
+succeeds.
+
+> [!NOTE]
```
rjmholt: I would put this note immediately under the `Long Description`

sdwheeler: Moved

reference/7/Microsoft.PowerShell.Core/About/about_Pipeline_Chain_Operators.md
```diff
@@ -0,0 +1,294 @@
...
+[conditional processing symbols](/previous-versions/windows/it-pro/windows-xp/bb490954(v=technet.10)#using-multiple-commands-and-conditional-processing-symbols)
+in the Windows Command Shell (cmd.exe).
+
+The `&&` operator executes the right-hand pipeline, if the left-hand pipeline
+succeeded. Conversely, the `||` operator executes the right-hand pipeline if
+the left-hand pipeline failed.
+
+These operators use the `$?` and `$LASTEXITCODE` variables to determine if a
+pipeline failed. This allows you to use them with native command and not just
+with cmdlets or functions. For example, this could be used for conditional
```
rjmholt: I want to stay away from the specifics of the example this early in the doc. I'll give a better example below

reference/7/Microsoft.PowerShell.Core/About/about_Pipeline_Chain_Operators.md
```diff
@@ -0,0 +1,294 @@
...
+execution of a software deployment:
+
+```powershell
+npm run build && npm run deploy
```
rjmholt: A better example might be this:

```powershell
# Create a new SSH key pair and if successful, copy the public key to the clipboard
ssh-keygen -t rsa -b 2048 && Get-Content -Raw ~\.ssh\id_rsa.pub | clip
```

reference/7/Microsoft.PowerShell.Core/About/about_Pipeline_Chain_Operators.md
sdwheeler: See my updates

reference/7/Microsoft.PowerShell.Core/About/about_Pipeline_Chain_Operators.md
sdwheeler: See my updates

reference/7/Microsoft.PowerShell.Core/About/about_Pipeline_Chain_Operators.md
```diff
@@ -0,0 +1,291 @@
...
+expressions like `+` or `-and`, for example.
+
+`&&` and `||` have a lower precedence than piping (`|`) or redirection (`>`),
+but a higher precedence than job operators (`&`), assignment (`=`) or
+semicolons (`;`). This means that pipelines within a pipeline chain can be
+individually redirected, and that entire pipeline chains can be backgrounded,
+assigned to variables, or separated as statements.
+
+To use lower precedence syntax within a pipeline chain, consider the use of
+parentheses `(...)` or a subexpression `$(...)`.
```
rjmholt: I omitted a thing here about `&&` and `||` being left associative (so `cmd1 && cmd2 && cmd3` is the same as `[cmd1 && cmd2] && cmd3`), but wasn't sure if that's too technical here.

It's relevant if you do something like:

```powershell
Get-Item ./file.txt || Write-Error 'File not found' && Get-Content -Raw ./file.txt
```

The associativity means that `Get-Content` will be executed based on the result of `Get-Item || Write-Error`, so if `Get-Item` succeeds it is executed, but if it doesn't `Write-Error` runs, which by definition does not succeed, so `Get-Content` does not succeed.

sdwheeler: I think it would be good to a note about associativity. Can associativity be altered using paretheses or braces? For example:

```
(Get-Item ./file.txt || Write-Error 'File not found') && Get-Content -Raw ./file.txt
{Get-Item ./file.txt || Write-Error 'File not found'} && Get-Content -Raw ./file.txt
```

rjmholt: Parentheses currently don't work nicely because they always set `$?` to `$true`... That's always been the case unfortunately, although this feature has many people talking about changing it.

Unfortunately the dot-sourced scriptblock also suffers from this.

rjmholt: Ok, I've added a note about associativity, plus a bit about `(...)` etc setting `$?`

sdwheeler: Looks good. Thanks

PR #4899, opened 2019-10-07
Fix bullet list in NOTES section

prp1277:
Fixed markdown formatting from leaving an empty bullet point to using a blockquote


sdwheeler:
@prp1277 Thank your for your contribution. Our style is to use normal paragraphs in the NOTES section or to use bulleted lists. Unfortunately, as you found and tried to fix, there are many cases where there is a single list item followed by an empty bullet. 

Rather than using a block quote, please remove the empty bullet item. You can also convert the single item to a paragraph or leave it as a single bullet. Also, to accept this as a PR, please update the same article in each versioned folder.

PR #4907, opened 2019-10-08
Add Debian 8 to list of supported versions

perlun:
It was already present in the list further down; this merely adds it to the header.

Unsure about the "version(s)" info below - I honestly don't understand how that part works. (I can't see any separate branches for these versions, so maybe there's some other magic going on here... 😄) 

- [ ] Impacts 7 document
- [ ] Impacts 6 document
- [ ] Impacts 5.1 document
- [ ] Impacts 5.0 document
- [ ] Impacts 4.0 document
- [ ] Impacts 3.0 document


sdwheeler:
@perlun Thanks for the contribution. The version info does not apply to articles in the `docs-conceptual` folder tree. There is a version numbered folder for all of the cmdlet reference. There are no separate branches for versions at this time.

PR #5771, opened 2020-04-13
fix table layout in source and result

schuelermine:
formatting fix


sdwheeler:
Thanks for the contribution, but the table has to be formatted this way because of limitations in our publishing system. Please see https://docs.microsoft.com/powershell/scripting/community/contributing/editing-cmdlet-ref#formatting-about_-files

PR #5937, opened 2020-05-15
Update PSReadLine docs

daxian-dbw:
Update PSReadLine docs with the most recent changes.
This is a follow-up PR to #5931


sdwheeler:
@daxian-dbw Thanks for the updates

PR #6128, opened 2020-06-16
add calls to Get-TargetResource

gforceg:
Make it clear that Get-TargetResource is often called by Set-TargetResource and Test-TargetResource.

# PR Summary

It wasn't obvious to me from this doc how to implement `Test-ResourceTarget` and I found out by looking at a few DSC Resources on github that `Get-ResourceTarget` is often used to drive `Set-ResourceTarget` and for comparison in `Test-ResourceTarget`. I think adding this to the page gives the user a clue 

## PR Context

Select the type(s) of documents being changed.

- [ ] Version 7.1 preview content
- [ ] Version 7.0 content
- [ ] Version 6 content
- [X] Version 5.1 content
(I do not know what other versions this affects, I am in an archaic development environment :D)

**Conceptual content**
- [ ] Fundamental conceptual articles
- [X] Script sample articles
- [X] DSC articles
- [ ] Gallery articles
- [ ] JEA articles
- [ ] WMF articles
- [ ] SDK articles
- [ ] Community content

## PR Checklist

### DIFF
--- a/reference/docs-conceptual/dsc/resources/authoringResourceMOF.md
+++ b/reference/docs-conceptual/dsc/resources/authoringResourceMOF.md
@@ -1,5 +1,5 @@
 ---
-ms.date:  06/12/2017
+ms.date:  06/16/2020
 keywords:  dsc,powershell,configuration,setup
 title:  Writing a custom DSC resource with MOF
 ---
@@ -154,6 +154,9 @@ function Set-TargetResource
         [string[]]$Protocol
     )
 
+    <# Get the current state #>
+    $currentState = Get-TargetResource -Ensure $Ensure -Name $Name -PhysicalPath $PhysicalPath -State $State -ApplicationPool $ApplicationPool -BindingInfo $BindingInfo -Protocol $Protocol
+    
     <# If Ensure is set to "Present" and the website specified in the mandatory input parameters does not exist, then create it using the specified parameter values #>
     <# Else, if Ensure is set to "Present" and the website does exist, then update its properties to match the values provided in the non-mandatory parameter values #>
     <# Else, if Ensure is set to "Absent" and the website does not exist, then do nothing #>
@@ -198,6 +201,9 @@ $BindingData,
 $ApplicationPool
 )
 
+<# Get the current state #>
+$currentState = Get-TargetResource -Ensure $Ensure -Name $Name -PhysicalPath $PhysicalPath -State $State -ApplicationPool $ApplicationPool -BindingInfo $BindingInfo -Protocol $Protocol
+    
 #Write-Verbose "Use this cmdlet to deliver information about command processing."
 
 #Write-Debug "Use this cmdlet to write debug information while troubleshooting."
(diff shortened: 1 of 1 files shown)


mgreenegit:
Thanks @gforceg. Calling Get from Test has developed as a common community practice. I could see documenting it as a best practice to help new authors figure it out (I agree it is not intuitive). I don't see it as often for Set though?

I'd like to get feedback from the DSC Community committee to get a broader opinion.
@johlju 
@PlagueHO 
@gaelcolas
@kwirkykat

gforceg:
Thanks for the feedback @mgreenegit. I look forward to getting it right.
I admit I made this PR in a rush and without all the info. It's not very professional but it gives the reader a thread to pull on. ;)

PlagueHO:
Agree that calling `Get` from `Test` has become fairly common practice. It is used in `Set`, but much less often. It is used more in newer resources that have also adopted the `Test-DscParameterState` pattern.

My feeling is that it isn't a "best practice" to use `Get` in `Test` or `Set` because some resources don't need to use that pattern. But I'm fine with using it as an example of a "good design pattern".

Could also therefore use include `Test-DscParameterState` as an example of a "good design pattern" - however that requires `DscResource.Common` module.

Perhaps build up a few reference patterns, including `Test-DscParameterState` pattern.

johlju:
I agree with @PlagueHO that `Get` is called from both `Test` and `Set`, but less often from `Set`. 

But if  `Invoke-DscResource` is used more in the future (WPS 5.1, and PS 7 and beyond) that means that you can call `Set` directly without first passing `Test` first as the LCM does it. That means that resource need to be refactored in the future to test the state in `Set` too to not set all values regardless if they are in desired state or not (resource should only change values that is not in desired state and that should be enforced). For that reason there is yet another pattern that is used by some (newer) resources in ActiveDirectoryDsc and SqlServerDsc. That pattern introduces a `Compare`-function that both `Test` and `Set` calls that calls a function `Compare-ResourcePropertyState`.

So there are these patterns (at least in the resources in DSC Community):

- `Get`, `Set`, and `Test` are self-sustaining not sharing any code (bad design pattern, these are often refactored when there is time)
- `Test` is calling Get using the mandatory parameters. Set is self-sustaining and often not having any evaluation code but just setting all properties (that should be enforced). Most resources uses this pattern in one way or another - I think because of legacy, because of LCM behavior.
- `Test` is calling `Test-DscParameterState` which is passed the result from `Get` and returns true/false. In this case `Set` is (most often) doing the same as previous item. Many resources uses this pattern.
- Both `Test` and `Set` are calling `Compare-TargetResourceState` which first calls `Get` and then calls `Compare-ResourcePropertyState` with result from `Get` and returns a hashtable with metadata of each property (`ParameterName`, `Actual`, `Expected`, `InDesiredState`).

Of course there are flavors of these patterns.

I'm trying to use last pattern when building new resources. An example resource is this one
https://github.com/dsccommunity/SqlServerDsc/blob/743be90fc9076b53d9de18a1f3d42f0e0f2d7c74/source/DSCResources/DSC_SqlProtocol/DSC_SqlProtocol.psm1#L234-L237

chasewilson:
@gforceg Thanks for the submission and @mgreenegit @PlagueHO @johlju, thanks for the input. 
For right now, @gforceg would you be able to remove the `Get-TargetResource` reference in the `Set-TeargetResource` function? 

I've opened a new issue to address some of the design patterns referenced above but would love to still use what you have here?

PR #6151, opened 2020-06-18
update to compare-object doc

jszabo98:
Trying to clear up confusion on how compare-object works.  The properties of objects will not be looked at, unless the -property parameter is used.

# PR Summary

## PR Context

Select the type(s) of documents being changed.

- [x] Version 7.1 preview content
- [x] Version 7.0 content
- [x] Version 6 content
- [x] Version 5.1 content

**Conceptual content**
- [ ] Fundamental conceptual articles
- [ ] Script sample articles
- [ ] DSC articles
- [ ] Gallery articles
- [ ] JEA articles
- [ ] WMF articles
- [ ] SDK articles
- [ ] Community content

## PR Checklist

- [ ] PR has a meaningful title
- [ ] PR is targeted at the _staging_ branch
- [ ] This PR is ready to merge and is not **Work in Progress**

### DIFF
--- a/reference/7.0/Microsoft.PowerShell.Utility/Compare-Object.md
+++ b/reference/7.0/Microsoft.PowerShell.Utility/Compare-Object.md
@@ -29,7 +29,7 @@ Compare-Object [-ReferenceObject] <PSObject[]> [-DifferenceObject] <PSObject[]>
 The `Compare-Object` cmdlet compares two sets of objects. One set of objects is the **reference**,
 and the other set of objects is the **difference**.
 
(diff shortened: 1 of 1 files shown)


sdwheeler:
@jszabo98 I took a closer look at the article. And I can see some confusing language related to Property comparisons. I did an extensive review of the source code with the Dev team. 

Compare-Object to not do any sorting. However, it can appear that way when there is a sufficiently large SyncWindow. The SyncWindows defaults to `[int]::maxvalue`, so has the same affect as sorting with anything smaller of that size. Also, the comparison logic is more complex than just "convert to string and compare".  The cmdlet uses the `Equals()` method defined here: https://github.com/PowerShell/PowerShell/blob/ac2b1cfc278575c45e054d9620f9ede2f9b5eac0/src/Microsoft.PowerShell.Commands.Utility/commands/utility/ObjectCommandComparer.cs#L73-L107

If the object implements IComparable, then we call `LanguagePrimitives.TryCompare()` tries to compare the whole objects. And if the objects are different types, it tries type coercion before comparing. If it is not IComparable, then we call `ToString()` and compare the string results.

But I can't merge this PR. But I am working on an update to this article.

PR #6457, opened 2020-08-08
Parsing: Commands in expression mode and other clarifications

yecril71pl:
# PR Summary

There are more things that are parsed in expression mode: not just strings and numbers but also unary operators, variable references and commands.
`$a = 2+2` is a statement, not a command, so I removed it from the examples.
I provided some more information about special characters and added a few additional illustrative examples.
I changed `echo(2+2)` to point out that tokenisation is not what the reader thinks it is.
I replaced `Write-Output` with `echo` so as not to distract the reader.

## PR Context

**Conceptual content**
- [ ] Overview and Install
- [ ] Learning PowerShell
  - [ ] PowerShell 101
  - [ ] Deep dives
  - [ ] Remoting
- [ ] Windows PowerShell
- [ ] DSC articles
- [ ] Community resources
- [ ] Sample scripts
- [ ] Gallery articles
  - [ ] Legacy SDK

- [X] Version 7.0 content
- [X] Version 6 content
- [X] Version 5.1 content

## PR Checklist

- [X] All relevant versions updated

### DIFF
--- a/reference/7.1/Microsoft.PowerShell.Core/About/about_Parsing.md
+++ b/reference/7.1/Microsoft.PowerShell.Core/About/about_Parsing.md
@@ -45,7 +45,7 @@ or in argument mode:
 
 The following table provides several examples of values processed in
 expression mode and argument mode and the evaluation of those
-values.
+values.  We assume that the value of the variable `a` is `4`.
 
 |       Example        |    Mode    |      Result       |
 | -------------------- | ---------- | ----------------- |
@@ -55,6 +55,9 @@ values.
 | `echo $a`            | Expression | 4 (integer)       |
 | `echo $-`            | Argument   | "$-" (string)     |
 | `echo $a/H`          | Argument   | "4/H" (string)    |
+| `echo a$a`           | Argument   | "a4" (string)     |
+| `echo a'$a'`         | Argument   | "a$a" (string)    |
+| `echo a"$a"`         | Argument   | "a4" (string)     |
 | `` `2+2``            | Expression | "2+2" (command)   |
 | `$a`                 | Expression | 4 (integer)       |
 | `$a/H`               | Expression | error             |
--- a/reference/7.1/Microsoft.PowerShell.Core/About/about_Parsing.md
+++ b/reference/7.1/Microsoft.PowerShell.Core/About/about_Parsing.md
@@ -19,29 +19,73 @@ interpret each token.
 
 For example, if you type:
 
-`Write-Host book`
+```powershell
+Write-Host book
+```
 
 PowerShell breaks the command into two tokens, `Write-Host` and
-`book`, and interprets each token independently.
+`book`, and interprets each token independently using one of two 
+major parsing modes: expression mode and argument mode.
 
-When processing a command, the PowerShell parser operates in expression mode
-or in argument mode:
+<!--
+01234567890123456789012345678901234567890123456789012345678901234567890123456789
+-->
 
-- In expression mode, character string values must be contained in
-  quotation marks. Numbers not enclosed in quotation marks are treated as
-  numerical values (rather than as a series of characters).
-  Initial unary operators, like `.`, `+` or `[int]`,
-  introduce the corresponding operation.
-  Variable references are evaluated. Splatting is forbidden.
-  Anything else will be treated as a command to be invoked.
+### Expression mode
 
-- In argument mode, each value is treated as an expandable string unless it
-  begins with one of the following special characters: dollar sign (`$`), at
-  sign (`@`), single quotation mark (`'`), double quotation mark (`"`), or a
-  parenthesis (`(`, `)`).  If preceded by one of these characters,
-  the value is treated as a value expression,
-  except for `$`, which is treated like an ordinary character
-  if not part of a valid variable reference.
+Expression mode is intended for combining expressions, 
+required for value manipulation in a scripting language.
+Expressions are representations of values in PowerShell syntax, 
+and can be simple or composite, for example:
+
+Literal expressions are direct representations of their values: 
+
+```powershell
+'hello', 32, $true
+```
+
+Variable expressions carry the value of the variable they reference: 
+
+```powershell
+$x, $script:path
+```
+Operators combine other expressions for evaluation: 
+
+```powershell
+- 12, -not $Quiet, 3 + 7, $input.Length -gt 1
+```
+
+* *Character string literals* must be contained in quotation marks. 
+* *Numbers* are treated as numerical values rather than as a series of characters 
+(unless escaped).
+* *Operators*, including unary operators like `-` and `-not` 
+and binary operators like `+` and `-gt`, are interpreted as operators 
+and apply their respective operations on their arguments (operands).
+* *Attribute and conversion expressions* are parsed as expressions 
+and applied to subordinate expressions, e.g. `[int] '7'`.
+* *Variable references* are evaluated to their values 
+but *splatting* (i.e. pasting prefilled parameter sets) is forbidden 
+and causes a parser error.
+* Anything else will be treated as a command to be invoked.
+
+### Argument mode
+
+When parsing, PowerShell first looks to interpret input as an expression.
+But when a command invocation is encountered, parsing continues in argument mode.
+
+Argument mode is designed for parsing arguments and parameters for commands 
+in a shell environment.  All input is treated as an expandable string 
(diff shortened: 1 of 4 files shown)


sdwheeler:
@rjmholt Can you review this PR?

reference/7.1/Microsoft.PowerShell.Core/About/about_Parsing.md
```diff
@@ -30,27 +30,40 @@ or in argument mode:
 - In expression mode, character string values must be contained in
   quotation marks. Numbers not enclosed in quotation marks are treated as
   numerical values (rather than as a series of characters).
+  Initial unary operators, like `.`, `+` or `[int]`,
+  introduce the corresponding operation.
```
rjmholt, line 34: Here I think `.`, `+` and `[int]` probably aren't good examples of unary operators:

- `.` as the dot-sourcing operator doesn't really fall under the expression side of PS syntax, and beginning an expression-mode clause with `.` will put you in the command rule, meaning command mode. So it's probably going to confuse users reading this.
- `+` is quite an unusual unary operator to use. Something like `-` or `-not` would be more common
- `[int]` is not really a unary operator but an expression in its own right that can be applied to another expression to form a *conversion expression*

To honest, the way this section is currently structured doesn't really lend itself to a proper introduction of the concepts required in my view. We try to cheat our way past explaining what an "expression" is, to get straight to "expression mode".

I would try to put it more like:

> PowerShell has two major parsing modes: expression mode and argument mode.
>
> Expression mode is intended for combining expressions, required for value manipulation in a scripting language.
> Expressions are representations of values in PowerShell syntax, and can be simple or composite, for example:
>    - Literal expressions are direct representations of their values: `'hello'`, `32`, `$true`
>    - Variable expressions carry the value of the variable they reference: `$x`, `$script:path`
>    - Operators combine other expressions for evaluation: `- 12`, `-not $Quiet`, `3 + 7`, `$input.Length -gt 1`
>
> In expression mode:
>    - Character string literals must be contained in quotation marks. Numbers not enclosed in quotation marks are treated as numerical values rather than as a series of characters.
>    - Operators, including unary operators like  `-` and `-not`, and binary operators like `+` and `-gt`, are interpreted as operators and apply their respective operations on their arguments (operands).
>    - Attribute and conversion expressions are parsed as expressions and applied to subordinate expressions: `[int] '7'`
>    - Variables are evaluated to their values, but splatting causes a parser error.
>    - Anything else will be treated as a command to be invoked.
>
> When parsing, PowerShell first looks to interpret input as an expression.
> But when a command invocation is encountered, parsing continues in argument mode.
>
> Argument mode is designed for parsing arguments and parameters for commands in a shell environment, and has a bias toward strings.
> In argument mode, all input is treated as an expandable string unless it uses one of the following syntaxes:
>
>    - Dollar sign (`$`), which begins a variable reference (unless it begins an invalid variable name, in which case it is interpreted as part of the expandable string)
>    - Quotation marks (`'` and `"`), which begin string values
>    - Parentheses (`(` and `)`) and subexpressions (`$(...)`), which demarcate new expressions
>    - Braces (`{` and `}`), which denote a scriptblock
>    - The at sign (`@`), which begins expression syntaxes such as splatting (`@args`), arrays (`@(1, 2, 3)`) and hashtables (`@{ a = 1; b = 2 }`)

rjmholt: We could probably use more comparative examples in this article too like:

> | Syntax | Expression Mode Input | Expression Mode Result | Argument Mode Input | Argument Mode Result |
> | ----- | ----- | ----- | ----- | ----- |
> | 'hello' | 'hello' | `hello` (string) | `Write-Output 'hello'` | `hello` (string) |
> | `1+1` | `1+1` | `2` (integer) | `Write-Output 1+1` | `1+1` (string) |
> | `[int]'3'` | `[int]'3'` | `3` (integer) | `Write-Output [int]'3'` | `[int]'3'` (string) |
> | `Get-Date` | `Get-Date` | `Sunday, 9 August 2020 2:41:25 pm` (datetime value) | `Write-Output Get-Date` | `Get-Date` (string) |
> | `(1+1)` | `(1+1)` | `2` (integer) | `Write-Output (1+1)` | `2` (integer) |

reference/7.1/Microsoft.PowerShell.Core/About/about_Parsing.md
```diff
@@ -30,27 +30,40 @@ or in argument mode:
...
-| `Write-Output $a`    | Expression | 4 (integer)       |
-| `Write-Output $a/H`  | Argument   | "4/H" (string)    |
+| `echo 2+2`           | Argument   | "2+2" (string)    |
+| `echo(2+2)`          | Expression | 4 (integer)       |
+| `echo $a`            | Expression | 4 (integer)       |
+| `echo $-`            | Argument   | "$-" (string)     |
+| `echo $a/H`          | Argument   | "4/H" (string)    |
+| `echo a$a`           | Argument   | "a4" (string)     |
+| `echo a'$a'`         | Argument   | "a$a" (string)    |
+| `echo a"$a"`         | Argument   | "a4" (string)     |
```
rjmholt, line 60: I'm not sure `echo` is simpler or better here

reference/7.1/Microsoft.PowerShell.Core/About/about_Parsing.md
yecril71pl: > more comparative examples

The table is too wide for 80 columns, we would have to shorten things, e.g. using `echo`.

rjmholt: > The table is too wide for 80 columns, we would have to shorten things, e.g. using echo.

Oh, yeah, the table probably needs some whittling down. Certainly the columns could be improved I think. I'm hoping @sdwheeler might have some suggestions there.

My issue with `echo` is that it's an alias on some platforms and a native command on others. Generally in docs we try to stay away from aliases and platform-specific commands unless we need something particular about them as behavioural examples. In this case, I'd prefer not to conflate the questions around `echo` with our description of parsing modes. It's probably worth diversifying the examples somewhat though, so we could well have one example that uses `echo`.

sdwheeler: What I have had to do with other articles to convert the table to a bulleted list. This allows for more description but it is not as succinct and makes it harder to see the differences between the modes.

But, that is my recommendation. Reformat as a list.

sdwheeler: > I would try to put it more like:

@rjmholt I this the rewrite you provided. Let's use that.

yecril71pl: > My issue with `echo` is that it's an alias on some platforms and a native command on others. Generally in docs we try to stay away from aliases and platform-specific commands unless we need something particular about them as behavioural examples. In this case, I'd prefer not to conflate the questions around `echo` with our description of parsing modes. It's probably worth diversifying the examples somewhat though, so we could well have one example that uses `echo`.

My understanding is that `echo` is a built-in alias for all platforms and it wins over native commands, as all aliases do.  Also, what `echo` means and does is _irrelevant_ to parsing, we could use `qwerty` instead—but it has the advantage of being able to see the result.  This is not the same as seeing the actual parsed input, of course, but it is better than nothing.  While it is also possible to provide _that_ in the examples, it would certainly make the readers run for their lives, so let’s rather not do that 😉

rjmholt: > My understanding is that echo is a built-in alias for all platforms and it wins over native commands

Ah, you're quite right there.

I agree with your points, but basically my thinking is that `Write-Output` is the canonical identity primitive in PowerShell. Docs should require a minimum of prerequisite knowledge to understand any given concept, and should also be consistent in examples when possible, and `Write-Output` fits those requirements. @sdwheeler is the ultimate arbiter of what's preferred there though.

> But, that is my recommendation. Reformat as a list.

Yeah, that makes sense. The key part to me is the side-by-side comparison though, which the table does nicely. If I had to choose, I would go for a smaller table rather than bullets. I just want a volley of examples to really hit home how argument mode basically sees everything as an expandable string

yecril71pl:
I think examples like `echo 002` or `echo .300` are too stupefying to be included.

PR #6513, opened 2020-08-18
Fix mistakes in the description of ArgumentList

yecril71pl:
# PR Summary

1. Fix obvious mistakes.
2. Improve code layout.
3. The code does not involve using an array *subexpression*, so remove the term.

## PR Context

**Conceptual content**
- [ ] Overview and Install
- [ ] Learning PowerShell
  - [ ] PowerShell 101
  - [ ] Deep dives
  - [ ] Remoting
- [ ] Windows PowerShell
- [ ] DSC articles
- [ ] Community resources
- [ ] Sample scripts
- [ ] Gallery articles
  - [ ] Legacy SDK

- [X] Version 7.0 content
- [X] Version 6 content
- [X] Version 5.1 content

## PR Checklist

- [X] All relevant versions updated

### DIFF
--- a/reference/7.1/Microsoft.PowerShell.Core/About/about_Splatting.md
+++ b/reference/7.1/Microsoft.PowerShell.Core/About/about_Splatting.md
@@ -123,18 +123,21 @@ parameter values to a script block that is executed by the cmdlet. The
 script block. PowerShell is effectively using array splatting to bind the
 values to the parameters of the script block. When using **ArgumentList**, if
 you need to pass an array as a single object bound to a single parameter, you
-must wrap the array in an array subexpression.
+must wrap the array as the only element of another array.
 
 The following example has a script block that takes a single parameter that is
 an array of strings.
 
 ```powershell
 $array = 'Hello', 'World!'
-Invoke-Command -ScriptBlock { param([string[]]$words) $words -join ' '} -ArgumentList $array
+Invoke-Command -ScriptBlock {
+  param([string[]]$words) $words -join ' '
+  } -ArgumentList $array
 ```
-
-In this example, only the first item in `$array` is passed to the script script
-block.
+<!--
+01234567890123456789012345678901234567890123456789012345678901234567890123456789
+-->
+In this example, only the first item in `$array` is passed to the script block.
 
 ```Output
 Hello
@@ -142,10 +145,12 @@ Hello
 
 ```powershell
 $array = 'Hello', 'World!'
-Invoke-Command -ScriptBlock { param([string[]]$words) $words -join ' '} -ArgumentList (,$array)
+Invoke-Command -ScriptBlock {
+  param([string[]]$words) $words -join ' '
+} -ArgumentList (,$array)
 ```
 
-In this example, `$array` is wrapped in an array subexpress so that the entire
+In this example, `$array` is wrapped in an array so that the entire
 array is passed to the script block as a single object.
 
 ```Output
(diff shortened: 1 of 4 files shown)


reference/7.1/Microsoft.PowerShell.Core/About/about_Splatting.md
```diff
@@ -123,29 +123,34 @@ parameter values to a script block that is executed by the cmdlet. The
...
 $array = 'Hello', 'World!'
-Invoke-Command -ScriptBlock { param([string[]]$words) $words -join ' '} -ArgumentList $array
+Invoke-Command -ScriptBlock {
+  param([string[]]$words) $words -join ' '
+  } -ArgumentList $array
 ```
-
-In this example, only the first item in `$array` is passed to the script script
-block.
+<!--
```
chasewilson, line 137: I think this was an accident

reference/7.1/Microsoft.PowerShell.Core/About/about_Splatting.md
```diff
@@ -123,29 +123,34 @@ parameter values to a script block that is executed by the cmdlet. The
...
 
 ```powershell
 $array = 'Hello', 'World!'
-Invoke-Command -ScriptBlock { param([string[]]$words) $words -join ' '} -ArgumentList (,$array)
+Invoke-Command -ScriptBlock {
+  param([string[]]$words) $words -join ' '
+} -ArgumentList (,$array)
 ```
 
-In this example, `$array` is wrapped in an array subexpress so that the entire
```
chasewilson, line 148: I don't this needs to be corrected (other than the typo in expression.). IThe new phrassing makes it look as if you construct an array and pass it, it will work. However, passing it in like this will cause an error `-ArgumentList ,$array`

This will normally create an array and be able to be used as such but that's not the case here. It has to wrapped in an aray sub expression to work.

chasewilson: This also goes for the same comment up top

reference/7.1/Microsoft.PowerShell.Core/About/about_Splatting.md
yecril71pl: This is a ruler.  It helps to set the correct width of the browser window.

reference/7.1/Microsoft.PowerShell.Core/About/about_Splatting.md
yecril71pl: The example given does not use an array subexpression. Your syntax is unsupported and causes a parse error because `,` cannot start an argument.

chasewilson: The Parentheses make a subexpression though. You are right you could do it with the null but with the parentheses, the comma in fromnt of the item contstructs and array. It's not the parentheses that  construct the array, they just run the sub expression and allow the output to be used and in this case, it outputs a single item array. 

You  could make the case to wrap it in the array subexpression operator (`@()`) as well and it will do the same thing. but either way, it's not wrapped in an array.

yecril71pl: The term *array subexpression* has a specific meaning related to the array subexpression operator.  Within this context, the statement that an array subexpression is required is simply not true.  Within the general context, when you assume that an array subexpression is a any subexpression that returns an array, this is not true either, because `($array)` is an array subexpression too.  
In this case, the parentheses serve as a grouping operator.  They do not introduce a subexpression, they constitute the whole expression.  While technically every expression is a subexpression of itself, I consider this term misleading in this context because it shifts the emphasis from what it is (an array within an array) to how it is written (and there is a number of ways to achieve this effect).

chasewilson: Hmm I can see your point and can see now that it could be interpreted too broadly. Your wording doesn't complicate it . 

Thanks for your patience. looks good then. Can you add these changes to the other versions and I will merge.

chasewilson: I agree that array subexpression could be too close to the operator. However in this context and the way that it reads, it's pretty clear that it's not

PR #6518, opened 2020-08-19
About -- and escaping parameter names

yecril71pl:
# PR Summary

How to pass parameter names as parameter values.
Also what happens to unrecognised named parameters.

## PR Context

**Conceptual content**
- [ ] Overview and Install
- [ ] Learning PowerShell
  - [ ] PowerShell 101
  - [ ] Deep dives
  - [ ] Remoting
- [ ] Windows PowerShell
- [ ] DSC articles
- [ ] Community resources
- [ ] Sample scripts
- [ ] Gallery articles
  - [ ] Legacy SDK

- [ ] Version 7.0 content
- [ ] Version 6 content
- [ ] Version 5.1 content

## PR Checklist

### DIFF
--- a/reference/7.1/Microsoft.PowerShell.Core/About/about_Parameters.md
+++ b/reference/7.1/Microsoft.PowerShell.Core/About/about_Parameters.md
@@ -35,10 +35,11 @@ Some parameters do not require or accept a parameter value. Other parameters
 require a value, but do not require the parameter name in the command.
 Named parameters passed to a command that does not expect them become positional
 parameters (strings) followed by their given values, if any.
-If a parameter name is abbreviated, like `-I`, PowerShell will pick the parameter
-that begins with this prefix, like `-InputObject`, provided that there is only
-one such parameter.  Because a future version may add more parameters starting
-with the same prefix, using this feature in scripts is not recommended.
+If a parameter name is abbreviated, like `-I`, PowerShell picks the parameter
+that begins with this prefix, like `-InputObject`. This only works if there is
+a single parameter starting with the specified prefix. Using this feature
+in scripts is not recommended as future versions of commands could include new
+parameters that start with the same prefix.
 
 The type of parameters and the requirements for those parameters vary. To find
 information about the parameters of a command, use the `Get-Help` cmdlet. For
--- a/reference/7.1/Microsoft.PowerShell.Core/About/about_Parameters.md
+++ b/reference/7.1/Microsoft.PowerShell.Core/About/about_Parameters.md
@@ -30,7 +30,7 @@ This effect can be prevented by escaping the hyphen (``Write-Output `-I``)
 or specifying the string as a parameter value (`Write-Output -InputObject:-I`).
 The special parameter `--` switches off recognising parameters for the rest
 of the command (`Write-Output -- -I`).
-The parameter name and value can be separated by a space or a colon character.
+The parameter name and value can be separated by a space or a colon character (`:`).
 Some parameters do not require or accept a parameter value. Other parameters
 require a value, but do not require the parameter name in the command.
 Named parameters passed to a command that does not declare them as parameters
--- a/reference/7.1/Microsoft.PowerShell.Core/About/about_Parameters.md
+++ b/reference/7.1/Microsoft.PowerShell.Core/About/about_Parameters.md
@@ -24,7 +24,7 @@ parameters follow the command name and have the following form:
 <!--
 01234567890123456789012345678901234567890123456789012345678901234567890123456789
 -->
-The name of the parameter is preceded by a hyphen (-), which signals to
+The name of the parameter is preceded by a hyphen (`-`), which signals to
 PowerShell that the word following the hyphen is a parameter name.
 This effect can be switched off by escaping the hyphen (``Write-Output `-I``)
 or specifying the string as a parameter value (`Write-Output -InputObject:-I`).
@@ -256,4 +256,3 @@ information about common parameters, see [about_CommonParameters](about_CommonPa
 [about_Pipelines](about_Pipelines.md)
 
 [about_Wildcards](about_Wildcards.md)
-
--- a/reference/7.1/Microsoft.PowerShell.Core/About/about_Parameters.md
+++ b/reference/7.1/Microsoft.PowerShell.Core/About/about_Parameters.md
@@ -27,7 +27,7 @@ parameters follow the command name and have the following form:
 The name of the parameter is preceded by a hyphen (-), which signals to
 PowerShell that the word following the hyphen is a parameter name.
 This effect can be switched off by escaping the hyphen (``Write-Output `-I``)
-or specifying the string as a parameter value (`Write-Output -I:-I`).
+or specifying the string as a parameter value (`Write-Output -InputObject:-I`).
 The special parameter `--` switches off recognising parameters for the rest
 of the command (`Write-Output -- -I`).
 The parameter name and value can be separated by a space or a colon character.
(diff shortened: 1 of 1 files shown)


reference/7.1/Microsoft.PowerShell.Core/About/about_Parameters.md
```diff
@@ -21,11 +21,24 @@ parameters follow the command name and have the following form:
...
+The name of the parameter is preceded by a hyphen (-), which signals to
+PowerShell that the word following the hyphen is a parameter name.
+This effect can be switched off by escaping the hyphen (``Write-Output `-I``)
+or specifying the string as a parameter value (`Write-Output -InputObject:-I`).
+The special parameter `--` switches off recognising parameters for the rest
+of the command (`Write-Output -- -I`).
+The parameter name and value can be separated by a space or a colon character.
+Some parameters do not require or accept a parameter value. Other parameters
 require a value, but do not require the parameter name in the command.
+Named parameters passed to a command that does not expect them become positional
```
chasewilson, line 36: I'm sorry if I'm missing something in the section, can you give an example of a command like this? It seems strance that a named param would become positional. I could be just misunderstanding something.

reference/7.1/Microsoft.PowerShell.Core/About/about_Parameters.md
```diff
@@ -21,11 +21,24 @@ parameters follow the command name and have the following form:
...
+parameters (strings) followed by their given values, if any.
+If a parameter name is abbreviated, like `-I`, PowerShell will pick the parameter
+that begins with this prefix, like `-InputObject`, provided that there is only
```
chasewilson, line 39: can we end the sentence after `-InputObject`? These clasuses could be two separate sentences and runs on as it is right now. So, it could read something like this:

If a parameter name is abbreviated, like `-I`, PowerShell picks the parameter that begins with this prefix, like `-InputObject`. This only works if there is a single parameter starting with the specified prefix. Using this feature in scripts is not recommended as future versions of commands could include new parameters that start with the same prefix.

reference/7.1/Microsoft.PowerShell.Core/About/about_Parameters.md
yecril71pl:
```powershell
ECHO -A:B
```

chasewilson: Hmm ok, what you mean. It's confusing because it reads like you are talking about named parameters that exist for the cmdlet and those would be interpreted as strings for some reason. Any possiblity of re wording to make it a little clearer?

reference/7.1/Microsoft.PowerShell.Core/About/about_Parameters.md
```diff
@@ -21,11 +21,25 @@ parameters follow the command name and have the following form:
...
+The name of the parameter is preceded by a hyphen (`-`), which signals to
+PowerShell that the word following the hyphen is a parameter name.
+This effect can be prevented by escaping the hyphen (``Write-Output `-I``)
+or specifying the string as a parameter value (`Write-Output -InputObject:-I`).
+The special parameter `--` switches off recognising parameters for the rest
+of the command (`Write-Output -- -I`).
+The parameter name and value can be separated by a space or a colon character.
+Some parameters do not require or accept a parameter value. Other parameters
 require a value, but do not require the parameter name in the command.
+Named parameters passed to a command that does not declare them as parameters
```
chasewilson, line 36: I'm not sure about this. While it does work for Write-Output, if you try and do somethin similar with Get-Item, You get an Error saying  A parameter cannot be found.... 

The -Path param is a positional param for Get-Item but there error returned isn't about the path being able to be found, the expected error for that. This behavior is much more common in my experience for the parameters. So is there a way to specify it more or is it something that shouldn't be brounght up here as it's uncommon behavior?

sdwheeler: We should not be using `Write-Object` for examples on how parameters work. It is a special case.

The **InputObject** parameter has the attribute `ValueFromRemainingArguments = true`. That means that any string that does not match a defined parameter is treated as a value that is assigned to **InputObject**. If we document anything about `Write-Object` in this article, we need to document the resulting behavior of `ValueFromRemainingArguments`. It should be a separate section of the document with its own examples.

yecril71pl: This behaviour is not that uncommon, given that any simple function and any external application will get named parameters that way.  `ValueFromRemainingArguments` is a special provision to relax requirements of strict cmdlets.  I am more worried about user functions getting unexpected parameters that way than about users getting errors from passing unexpected parameters to strict cmdlets; the latter is expected, the former—not so.  In other words, if a parameter specification is used to set a parameter value within a function, an unexpected parameter is expected to do nothing or error out.  It is not expected to be passed as a string, which is why I think it should be documented.

chasewilson: If my testing is correct, this is pretty uncommon. Less than 5 percent of cmdlets that even have the `ValueFromRemainingArguments` parameter attribute actually accept it as a string or string array this way. This includes custom functions, Imported modules and all standard modules.  18 commands work and over 400 don't. 

Even in a simple function (nothing defining it as an advanced function. No cmdletBinding and not parameter attributes), while the bad parameter doesn't throw an error, it also doesn't seem to be accessible in the function. Not in the input variable or in any form. If you add a `Position` parameter attribute (making it act as an advanced function), the defualt behavior seems be to throw the "A parameter cannot be found... " error.

I don't see how this adds to this article for someone trying to understand the normal behavior of parameters. If there is something in about_Advanced_Functions or something similar that can be added to clarify the behavior that you're talking about, (which, I agree, does exist, just not to the extent of "normal" behavior) I'm open to that. But I don't see it adding value in this article.

yecril71pl: This is not about cmdlets, it is about user functions and applications.  There are infinitely many more of those than of cmdlets.  The extraneous arguments are perfectly accessible: 
```powershell
& { $ARGS } -A:A
CMD /CECHO -A:A
```
If PowerShell is a shell, its primary reason for existence is to invoke applications.  PowerShell functions are not cmdlets by default.  Thus this is the normal behaviour while cmdlet behaviour is special behaviour.

rjmholt: I think if we're documenting things in enough detail that the differences in parameter binding for different command types are appearing, then we need to acknowledge that different kinds of commands bind parameters differently.

There are commonalities that we can introduce initially as introductory concepts, but cmdlets, functions/scriptblocks and native commands all have their own processors and parameter binders, so different concepts apply.

The way a command handles different parameters simply depends on what kind of command it is.

I think trying to conflate all kinds of commands together in a single description will ultimately just prove confusing, since `named parameters passed to a command that does not declare them as parameters become positional parameters` may be true for e.g. native commands, but `If a parameter name is abbreviated... PowerShell picks the parameter that begins with that prefix` is not true of native commands at all. That may seem clear to us, but I suspect such a line could create such an expectation in users.

My recommendation is to give a broad introduction of parameters and how they are generally used/passed to commands, and then give more detailed explanations in separate specialised paragraphs pertaining to particular kinds of commands.

sdwheeler:
I don't think we can take this PR. If you think there is something missing from the docs, please open an issue so we can discuss. Also, the side effect of using `--` is unexpected, as you point out, and seems like a bug that should be filed.

PR #6687, opened 2020-09-29
Fixes #6679 - Updates Out-String

### LINKED ISSUE #6679: Various corrections and improvements to the Out-String topic
Cmdlet synopsis:

> Sends objects to the host as a series of strings.

The cmdlet sends output to the _pipeline_. There is nothing host-specific about it; as with any pipeline-outputting cmdlet, the host is only involved as the _default_ output target.

By default, a _single_, multi-line string is output; only with `-Stream` do you get a series.

The purpose of `Out-String` is to get the same representation that you see in the terminal, via PowerShell's formatting system, as a _string object_.

---

"Description" section:

> The Out-String cmdlet converts the objects that PowerShell manages into an array of strings. 

"that PowerShell manages" is confusing - I think something like "its input objects"  would suffice.

As stated, a _single_ string is output by default; only with `-Stream` and when captured in a variable do you get an _array_ of strings.

> By default, Out-String accumulates the strings and returns them as a single string, but you can use the Stream parameter to direct Out-String to return one string at a time. 

Better: "one _line_ at a time"

---

"Examples" section:

* Example 1 is confusing, because there is no reason to combine `Get-Content` - which outputs _strings_ to begin with - with `Out-String` (if you want to get the file content as a _single_ string, use `-Raw` instead).
  * The "Note" paragraph adds to the confusion: why would you want to compare a direct `Get-Content` call with an `Out-String` mediated one with respect to _properties_? True, the provider-added ETS properties are lost when you duplicate a string, but that is neither important nor specific to `Out-String` use.

* Example 2 should be replaced with something that shows a more meaningful use of the `-Width` parameter, namely one based on (implicit) `Format-Table` use where `-Width` can be used to prevent _truncation_:

```powershell
# Assuming your terminal is less than 231 chars. wide (the extra 31 chars. are for the first column, 'Name'),
# the output would be truncated without the -Width argument.
@{ foo = 'x' * 200 } | Out-String -Width 250
```

As a (hopefully) moot aside, the "Note" paragraph for the existing example is incorrect: since `-Stream` isn't used, there is no _array_ to inspect.

* Example 3:

> to convert each object to a string

No, there is no 1-to-1 correspondence between input objects and output strings.
Instead, whatever - typically multiline - representation results from the formatting is sent _line by line_.

---

"Parameters" section:

* `-Stream`:

> Indicates that the cmdlet sends a separate string for each object

See comments re example 3 above.

* `-Width`:

The current topic reads:

> Any additional characters are truncated, not wrapped

PR #6677, which is currently pending publication, changed that to:

> Any additional characters are wrapped to the next line.

Both versions are _half_ correct:

Whether wrapping or truncation occurs depends on the (possibly implicitly) applied `Format-*` cmdlet:

* _Truncation_ (with `…`)  only ever applies to (implied) `Format-Table` and `Format-Wide` formatting.

  * That's why I proposed  `@{ foo = 'x' * 200 } | Out-String -Width 250` for example 2 above, because it implicitly uses `Format-Table`, and the `-Width` argument indeed prevents truncation.

* _Line-wrapping_ applies to:
    * _all other_ `Format-*` cmdlets, notably the implicitly used `Format-List` in example 2.
    * _out-of-band_ formatted objects: `[string]` (printed as-is) and .NET primitive types and similar property-less value types such as `[decimal]` and `[bigint]` (printed as-is via `.ToString()`)

> The default value for the PowerShell console is 80 characters.

This is ambiguous, because it could be interpreted as `80` applying in _all_ console windows, irrespective of their actual width.

Better to say: In terminal (console) windows, the current window width is used as the default value.
PowerShell console windows default to a width of 80 characters on installation.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: fafeac93-aee7-9986-910f-8ede69d07d27
* Version Independent ID: 1925a288-f947-4a6e-cf91-8b106ef9e493
* Content: [Out-String (Microsoft.PowerShell.Utility) - PowerShell](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/out-string?view=powershell-7.1&WT.mc_id=ps-gethelp)
* Content Source: [reference/7.1/Microsoft.PowerShell.Utility/Out-String.md](https://github.com/MicrosoftDocs/PowerShell-Docs/blob/staging/reference/7.1/Microsoft.PowerShell.Utility/Out-String.md)
* Product: **powershell**
* Technology: **powershell-cmdlets**
* GitHub Login: @JamesWTruher
* Microsoft Alias: **jimtru**

chasewilson:
# PR Summary

Multiple updates to `Out-String`

## PR Context

Fixes #6679 
Fixes [AB#1776689](https://dev.azure.com/mseng/677da0fb-b067-4f77-b89b-f32c12bb8617/_workitems/edit/1776689)

**Conceptual content**
- [ ] Overview and Install
- [ ] Learning PowerShell
  - [ ] PowerShell 101
  - [ ] Deep dives
  - [ ] Remoting
- [ ] Windows PowerShell
- [ ] DSC articles
- [ ] Community resources
- [ ] Sample scripts
- [ ] Gallery articles
  - [ ] Legacy SDK

- [x] Version 7.1 preview content
- [x] Version 7.0 content
- [x] Version 6 content
- [x] Version 5.1 content

## PR Checklist

- [x] I have read the [contributors guide][contrib] and followed the style and process guidelines
- [x] PR has a meaningful title
- [x] PR is targeted at the _staging_ branch
- [x] All relevant versions updated
- [x] Includes content related to issues and PRs - see [Closing issues using keywords][key].
- [x] This PR is ready to merge and is not **Work in Progress**

### DIFF
--- a/reference/5.1/Microsoft.PowerShell.Utility/Out-String.md
+++ b/reference/5.1/Microsoft.PowerShell.Utility/Out-String.md
@@ -3,7 +3,7 @@ external help file: Microsoft.PowerShell.Commands.Utility.dll-Help.xml
 keywords: powershell,cmdlet
 Locale: en-US
 Module Name: Microsoft.PowerShell.Utility
-ms.date: 09/25/2020
+ms.date: 09/29/2020
 online version: https://docs.microsoft.com/powershell/module/microsoft.powershell.utility/out-string?view=powershell-5.1&WT.mc_id=ps-gethelp
 schema: 2.0.0
 title: Out-String
@@ -40,6 +40,29 @@ strings.
 ```powershell
 $C = Get-Culture | Select-Object -Property *
 Out-String -InputObject $C -Width 100
+```
+
+```Output
+Parent                         : en
+LCID                           : 1033
+KeyboardLayoutId               : 1033
+Name                           : en-US
+IetfLanguageTag                : en-US
+DisplayName                    : English (United States)
+NativeName                     : English (United States)
+EnglishName                    : English (United States)
+TwoLetterISOLanguageName       : en
+ThreeLetterISOLanguageName     : eng
+ThreeLetterWindowsLanguageName : ENU
+CompareInfo                    : CompareInfo - en-US
+TextInfo                       : TextInfo - en-US
+IsNeutralCulture               : False
+CultureTypes                   : SpecificCultures, InstalledWin32Cultures
+NumberFormat                   : System.Globalization.NumberFormatInfo
+DateTimeFormat                 : System.Globalization.DateTimeFormatInfo                                                Calendar                       : System.Globalization.GregorianCalendar                                                 OptionalCalendars              : {System.Globalization.GregorianCalendar, System.Globalization.
+GregorianCalendar}
+UseUserOverride                : True
+IsReadOnly                     : True
 ```
 
 The `$C` variable stores a **Selected.System.Globalization.CultureInfo** object. The object is the
@@ -183,7 +206,7 @@ formatter for the specified display destination.
 
 ## RELATED LINKS
 
-[about_Formatting] (../Microsoft.PowerShell.Core/About/about_Format.ps1xml.md)
+[about_Formatting](../Microsoft.PowerShell.Core/About/about_Format.ps1xml.md)
 
 [Out-Default](../Microsoft.PowerShell.Core/Out-Default.md)
 
--- a/reference/5.1/Microsoft.PowerShell.Utility/Out-String.md
+++ b/reference/5.1/Microsoft.PowerShell.Utility/Out-String.md
@@ -183,6 +183,8 @@ formatter for the specified display destination.
 
 ## RELATED LINKS
 
+[about_Formatting] (../Microsoft.PowerShell.Core/About/about_Format.ps1xml.md)
+
 [Out-Default](../Microsoft.PowerShell.Core/Out-Default.md)
 
 [Out-File](Out-File.md)
--- a/reference/5.1/Microsoft.PowerShell.Utility/Out-String.md
+++ b/reference/5.1/Microsoft.PowerShell.Utility/Out-String.md
@@ -12,7 +12,7 @@ title: Out-String
 # Out-String
 
 ## SYNOPSIS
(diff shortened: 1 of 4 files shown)


reference/5.1/Microsoft.PowerShell.Utility/Out-String.md
```diff
@@ -24,35 +24,15 @@ Out-String [-Stream] [-Width <Int32>] [-InputObject <PSObject>] [<CommonParamete
...
->
-> `Get-Content -Path C:\Test\Testfile.txt | Select-Object -Property *`
->
-> `Get-Content -Path C:\Test\Testfile.txt | Out-String | Select-Object -Property *`
-
-### Example 2: Get the current culture and convert the data to strings
+### Example 1: Get the current culture and convert the data to strings
 
 This example gets the regional settings for the current user and converts the object data to
 strings.
```
sdwheeler, line 38: Add an output block for this example.

reference/5.1/Microsoft.PowerShell.Utility/Out-String.md
```diff
@@ -99,6 +79,27 @@ matches for the text **gcm**.
 > If you omit the **Stream** parameter, the command displays all the aliases because `Select-String`
 > finds the text **gcm** in the single string that `Out-String` returns.
 
+### Example 3: Use the Width parameter to prevent truncation.
+
+While most output from `Out-String` is wrapped to the next line if it's longer than the specified
+width, there are some scenarios where the output will be truncated due to the formatter cmdlet
+`Out-String` uses for formatting. You can prevent truncation by using the **Width** parameter.
```
sdwheeler, line 86:
```suggestion
While most output from `Out-String` is wrapped to the next line, there are scenarios where the
output is truncated by the formatting system before being passed to `Out-String`. You can avoid
truncation using the **Width** parameter.
```

sdwheeler (changes requested):

sdwheeler (approved):

PR #6735, opened 2020-10-09
Add description metadata to About files

sdwheeler:
Bulk update


sdwheeler:
Related to AB#1786853

PR #7797, opened 2021-07-13
Remove duplicate code example in about_For.md

Hrxn:
The "Support for multiple operations" section has two example paragraphs, one for `init` and one for `repeat`
The first example for-loop in the `repeat` part is the same example for-loop already demonstrated at the beginning of the `init` part.

# PR Summary

## PR Context

**Conceptual content**
- [ ] Overview and Install
- [ ] Learning PowerShell
  - [ ] PowerShell 101
  - [ ] Deep dives
  - [ ] Sample scripts
  - [ ] Remoting
- [ ] Windows PowerShell
- [ ] DSC articles
- [ ] Community resources
- [ ] Gallery articles
  - [ ] Language Spec
  - [ ] Legacy SDK

- [ ] Preview content
- [ ] Version 7.1 content
- [ ] Version 7.0 content
- [ ] Version 5.1 content

## PR Checklist

- [ ] PR has a meaningful title
- [ ] PR is targeted at the _staging_ branch
- [ ] This PR is ready to merge and is not **Work in Progress**

### DIFF
--- a/reference/7.0/Microsoft.PowerShell.Core/About/about_For.md
+++ b/reference/7.0/Microsoft.PowerShell.Core/About/about_For.md
@@ -82,13 +82,6 @@ The following syntaxes are supported for multiple assignment operations in the
 **Repeat** statement:
 
 ```powershell
-# Comma separated assignment expressions.
-for (($i = 0), ($j = 0); $i -lt 10; $i++)
-{
-    "`$i:$i"
-    "`$j:$j"
-}
-
 # Comma separated assignment expressions enclosed in parenthesis.
 for (($i = 0), ($j = 0); $i -lt 10; ($i++), ($j++))
 {


chasewilson:
Hey @Hrxn, Thanks for your contribution here!

What it looks like happened is that the example wasn't updated to show the multiple operations in the **Repeat** section. Instead of removing the example entirely, would you be able to update the example to show multiple operations using comma separation in the **Repeat** section:

```powershell
for (($i = 0), ($j = 0); $i -lt 10; $i++, $j++)
{
    "`$i:$i"
    "`$j:$j"
}.

This change also needs to be made across all versions of the doc. 

Thanks again for catching this and your contribution!

**Comment by Hrxn** [comment]
> What it looks like happened is that the example wasn't updated to show the multiple operations in the **Repeat** section. Instead of removing the example entirely, would you be able to update the example to show multiple operations using comma separation in the **Repeat** section[.]

Hey, not entirely sure what you mean here.
The next example in the **Repeat** section (which is now the first example with this commit) already shows what you describe here - multiple operators in parentheses in the `Repeat` statement.

In other words, the first example in the **Repeat** section is already based on the first example in the **Init** section, and correctly expands on it.

**Comment by chasewilson** [comment]
> Hey, not entirely sure what you mean here.
> The next example in the **Repeat** section (which is now the first example with this commit) already shows what you describe here - multiple operators in parentheses in the `Repeat` statement.
> 
> In other words, the first example in the **Repeat** section is already based on the first example in the **Init** section, and correctly expands on it.

In the example from the init section, it's talking about multiple actions taken there, while only 1 action is taken in the **Repeat** section. The example you are removing is supposed to show multiple actions taken in the **Repeat** but the example wasn't updated to adjust `$i` and `$j`. So in my example I gave in my earlier response now adjusts both `$i` and `$j`. Showing every way you can have multiple actions in **Repeat** including with and without parenthises. 

Does that make sense?

**Comment by Hrxn** [comment]
Yeah, I think I get what you mean. In the example you gave
```powershell
for (($i=0), ($j=0); $i -lt 10; $i++, $j++)
{
    "`$i:$i"
    "`$j:$j"
}
```

You use two assignment operations in parenthesis in the init section, and two increment operations in the repeat section - without any parenthesis.

I've just tried this here, and this is apparently something which *only* works in the repeat part, but not in the init part.
Reasonable enough to also mention this as syntax variant, but maybe the documentation should express this distinction more explicitly. I don't know, what do you think?

chasewilson:
> Reasonable enough to also mention this as a syntax variant, but maybe the documentation should express this distinction more explicitly. I don't know, what do you think?

I suppose it could. Currently, the example explainations are written as comments in each code example. W might be able to clarify this one specifically a bit more. E.g. `# Comma separated assignment expressions in the Repeat portion.`. Or we could take a look at moving the example descriptions out of the code blocks?

chasewilson:
Hey @Hrxn, just following up with this issue. Do you have plans to make these changes?

sdwheeler:
This was fixed in #7826 - closing this PR.

PR #8819, opened 2022-05-16
Fixes #8817 - Add info about end-of-parameters token

### LINKED ISSUE #8817: Special treatment of -- (double dash) as end-of-parsing should be documented
### Prerequisites

- [X] Write a descriptive title.
- [X] Search the existing issues.
- [X] I am reporting the documentation problem for version of PowerShell I am using.

### Version

7.2

### Link to affected document

https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_parsing?view=powershell-7.2

### Description of the documentation error

It seems that `--` has a special meaning of "end-of-parameters" and therefore gets swallowed on its first occurrence. This seems to be the case when invoking CmdLets or external scripts. External applications seem not to be affected, but [there are indiciations](https://stackoverflow.com/questions/15780174/powershell-command-line-parameters-and) that in the past they were.

For example, `Write-Output a -- b -- c` will print `a` `b` `--` `c` (swallowing the first `--`) while `cmd.exe /c echo a -- b -- c` will print `a -- b -- c`.

I was unable to find any sort of documentation about this behavior. The only reference I found (cited in other posts about this issue on the Internet) is a third-party book ("Windows PowerShell in Action" by Bruce Payette). See [this Stack Overflow answer](https://stackoverflow.com/a/12198146/1871033).

The fact that this is not well-known, not documented and apparently also version-dependent makes "mysterious" issues appear, and makes it hard to reliably predict the behavior of some commands. What led me to opening this issue was [this seemingly odd problem](https://stackoverflow.com/questions/72259748/scripts-arguments-starting-with-are-removed/) that turned out to be related to certain npm versions creating an `npm.ps1` file additionally to `npm.cmd` which would then behave differently when invoked as something like `npm run test -- --value 10` (and I later found that [others had the same issue in the past too](https://stackoverflow.com/questions/70090402/why-arent-npm-scripts-relaying-arguments-correctly-on-windows-powershell)).

(Note: There is also `--%`, which _is_ documented but seems to behave in yet another way...)

### Suggested fix

I'd suggest to add documentation about `--` to the "Parsing" docs page (and perhaps also to "Parameters"), including a comparison between `--` and `--%`, under what circumstances `--` does or doesn't have special behavior, how to work around it if it's causing a problem and perhaps also version differences, if any.

sdwheeler:
# PR Summary

Fixes AB#1949656 - Fixes #8817 - Add info about end-of-parameters token

## PR Context

Check the boxes below to indicate the content affected by this PR.

**Repository or docset configuration**
- [ ] Repo documentation and configuration (.git/.github/.vscode etc.)
- [ ] Docs build files (.openpublishing.* and build scripts)
- [ ] Docset configuration (docfx.json, mapping, bread, module folder)

**Conceptual documentation**
- [x] Files in docs-conceptual

When changing **cmdlet reference** or **about_ topics**, the changes should be copied to all
relevant versions. Check the boxes below to indicate the versions affected by this change.

- [x] Preview content
- [x] Version 7.2 content
- [x] Version 7.1 content
- [x] Version 7.0 content
- [x] Version 5.1 content

## PR Checklist

- [x] I have read the [contributors guide][contrib] and followed the style and process guidelines
- [x] PR has a meaningful title
- [x] PR is targeted at the _main_ branch
- [x] All relevant versions updated
- [x] Includes content related to issues and PRs - see [Closing issues using keywords][key].
- [x] This PR is ready to merge and is not **Work in Progress**. If the PR is work in progress,
  please add the prefix `WIP:` or `[WIP]` to the beginning of the title and remove the prefix when
  the PR is ready.

### DIFF
--- a/reference/5.1/Microsoft.PowerShell.Core/About/about_Parsing.md
+++ b/reference/5.1/Microsoft.PowerShell.Core/About/about_Parsing.md
@@ -1,7 +1,7 @@
 ---
 description: Describes how PowerShell parses commands.
 Locale: en-US
-ms.date: 09/07/2021
+ms.date: 05/16/2022
 online version: https://docs.microsoft.com/powershell/module/microsoft.powershell.core/about/about_parsing?view=powershell-5.1&WT.mc_id=ps-gethelp
 schema: 2.0.0
 title: about Parsing
@@ -200,10 +200,11 @@ prevent PowerShell from misinterpreting the parentheses.
 icacls X:\VMS /grant Dom\HVAdmin:`(CI`)`(OI`)F
 ```
 
-### The stop-parsing token
+### The stop-parsing and end-of-parameters tokens
 
-Beginning in PowerShell 3.0, you can use the stop-parsing token (`--%`) to
-stop PowerShell from interpreting input as PowerShell commands or expressions.
+Beginning in PowerShell 3.0, you can use the _stop-parsing_ (`--%`) and
+_end-of-parameters_ tokens (`--`) to stop PowerShell from interpreting input as
+PowerShell commands or expressions.
 
 > [!NOTE]
 > The stop-parsing token is only intended for use on Windows platforms.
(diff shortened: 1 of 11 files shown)


michaeltlombardi (approved):

CherryDT:
Hm, my testing showed that `--` is ignored when invoking an `Application` (and passed on without being swallowed), but with an `ExternalScript` it is interpreted. The new docs don't seem to clarify that point.

PR #8909, opened 2022-06-06
6/06 PM Publish

huypub:


GregoryShields:
I see that about_Case-Sensitivity was merged, but in the version I'm running (7.2.6) it does not exist as a Help command.

That is, when I run...
`Help about_Case-Sensitivity`
...the topic is not found by the Help System.

There is, however, a webpage for the topic:
[about_Case-Sensitivity](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_case-sensitivity?view=powershell-7.2)

Please forgive me if I am missing something because of my own ignorance.

PR #9130, opened 2022-08-17
Documenting RetryIntervalSec value

sdarwin:
Open to your feedback.  Based on testing the default value for RetryIntervalSec is not "None", it's 5 seconds. Here is sample output.

```
Invoke-WebRequest http://github.com/foo/bar.tar.gz -Outfile bar.tar.gz -MaximumRetryCount 4 -Verbose
VERBOSE: GET with 0-byte payload
VERBOSE: Retrying after interval of 5 seconds. Status code for previous attempt: InternalServerError
VERBOSE: Retrying after interval of 5 seconds. Status code for previous attempt: InternalServerError
VERBOSE: Retrying after interval of 5 seconds. Status code for previous attempt: InternalServerError
```

It's waiting 5 seconds, not 0 seconds.

Another question, the -Verbose flag seems helpful. When I was debugging this issue I searched on the page for the words "debug" and "verbose". There's no mention of those flags.   Could that at least be mentioned somewhere on this page, perhaps use it in one of the examples.

## PR Checklist

### DIFF
--- a/reference/7.0/Microsoft.PowerShell.Utility/Invoke-RestMethod.md
+++ b/reference/7.0/Microsoft.PowerShell.Utility/Invoke-RestMethod.md
@@ -2,7 +2,7 @@
 external help file: Microsoft.PowerShell.Commands.Utility.dll-Help.xml
 Locale: en-US
 Module Name: Microsoft.PowerShell.Utility
-ms.date: 08/08/2022
+ms.date: 08/17/2022
 online version: https://docs.microsoft.com/powershell/module/microsoft.powershell.utility/invoke-restmethod?view=powershell-7&WT.mc_id=ps-gethelp
 schema: 2.0.0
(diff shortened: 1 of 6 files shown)


sdwheeler:
@sdarwin The **Debug** and **Verbose** parameters are part of the **CommonParameters**. See [about_CommonParameters](https://docs.microsoft.com/powershell/module/microsoft.powershell.core/about/about_commonparameters).

They are listed in the doc source file, but there is a long standing bug that causes them to not be displayed on the webpage. We are working on a fix but it is still a ways out.

As for the default, I think that the value of **RetryIntervalSec** is not set in the cmdlet but it is using the default value of the underlying API (5 seconds). That is why is is listed as none. But we can change the docs. Thanks for submitting this.

sdwheeler (approved):

PR #9686, opened 2023-01-11
Fixes #9231 - Add NoRunspaceAffinity

### LINKED ISSUE #9231: Need to add doc for the new `NoRunspaceAffinity` attribute introduced to PowerShell class
### Prerequisites

- [X] **Descriptive Title:** Write the title for this issue as a short synopsis. If possible, provide context. For example, "Document new `Get-Foo` cmdlet" instead of "New cmdlet."

### PowerShell Version

7.3

### Summary

This should apply to 7.4 and above versions.

See the code PR: 
- https://github.com/PowerShell/PowerShell/pull/18138

Make PowerShell class not affiliate with Runspace when declaring the `NoRunspaceAffinity` attribute.

By default, a PowerShell class is affiliated with the Runspace where it's created. This makes it not work in a parallel workflow, such as `ForEach-Object -Parallel` -- method invocations on the class would be marshalled back to the Runspace it was created, which could cause corruption to the state of that Runsapce or a deadlock. See https://github.com/PowerShell/PowerShell/issues/4003 and https://github.com/PowerShell/PowerShell/issues/3651 for examples.

With this PR, if the attribute `NoRunspaceAffinity` is declared for a PowerShell class, then the PowerShell class will not be affiliated with a particular Runspace/SessionState, and its method invocation (both instance and static) will use the default Runspace of the running thread and its current session state.

Of course, once declaring the attribute to the class, the class implementation should not depend on any states of a particular session state, because there is no guarantee the dependencies would be available at runtime.

### Details

_No response_

### Proposed Content Type

About Topic

### Proposed Title

Should be an update to "about Class"

### Related Articles

_No response_

sdwheeler:
# PR Summary

Fixes [AB#58924](https://dev.azure.com/msft-skilling/cebd7ef5-4282-448b-9701-88c8637581b7/_workitems/edit/58924) - Fixes #9231 - Add NoRunspaceAffinity

## PR Checklist

### DIFF
--- a/reference/7.4/Microsoft.PowerShell.Core/About/about_Classes.md
+++ b/reference/7.4/Microsoft.PowerShell.Core/About/about_Classes.md
@@ -911,7 +911,7 @@ while ($true) {
 ```
 
 > [!NOTE]
-> This example runs in an infinite loop. Enter <kbd>Ctrl</kdb>+<kbd>C</kbd> to
+> This example runs in an infinite loop. Enter <kbd>Ctrl</kbd>+<kbd>C</kbd> to
 > stop the execution.
 
 ## Importing classes from a PowerShell module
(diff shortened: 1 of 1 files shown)


michaeltlombardi (dismissed):
My only nits are:

1. Should we define and/or link to runspaces at the start of the article?
2. Are we standardizing on capitalizing `Runspace` as a noun in the documentation generally?

PR #9754, opened 2023-01-25
(GH-9744) Fix about_Using examples

### LINKED ISSUE #9744: about_Using: example 3 is misleading
### Prerequisites

### Links

https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_using?view=powershell-5.1#example-3---load-classes-from-an-assembly

### Summary

Unlike what the example suggests, `using assembly` currently can _not_ be used to load assemblies with types that a `class` in the same script file derives from: `using assembly` is currently executed _at runtime_, whereas types used in `class` definitions must already have been loaded into the session _at parse time_.

While the example technically works, it only does so because the referenced assembly is _loaded by PowerShell itself_ at the start of every session. In other words: the `using assembly` statement isn't needed at all in this example, but if it _were_, the code would fail.

This is a well-known problem. A fix has been green-lit in 2017, but hasn't been implemented yet: see https://github.com/PowerShell/PowerShell/issues/3641#issuecomment-297151112

### Details

_No response_

### Suggested Fix

_No response_

michaeltlombardi:
# PR Summary

Prior to this change, examples 2 and 3 for about_Using were misleading. Example 2 implied that the classes from a script module could be namespaced. Example 3 implied that authors could call `using` to use classes from the assembly when defining a class in PowerShell, which actually raises a parsing exception.

This change reworks both examples to function without errors or incorrect implications.

- Resolves #9744
- Fixes [AB#60738](https://dev.azure.com/msft-skilling/cebd7ef5-4282-448b-9701-88c8637581b7/_workitems/edit/60738)

## PR Checklist

### DIFF
--- a/reference/5.1/Microsoft.PowerShell.Core/About/about_Using.md
+++ b/reference/5.1/Microsoft.PowerShell.Core/About/about_Using.md
@@ -1,7 +1,7 @@
 ---
 description: Allows you to indicate which namespaces are used in the session.
 Locale: en-US
-ms.date: 08/18/2022
+ms.date: 01/25/2023
 online version: https://learn.microsoft.com/powershell/module/microsoft.powershell.core/about/about_using?view=powershell-5.1&WT.mc_id=ps-gethelp
 schema: 2.0.0
 title: about Using
@@ -133,8 +133,8 @@ $hashfromstream.Hash.ToString()
 In this example, we have a PowerShell script module named **CardGames** that
 defines the following classes:
 
-- **CardGames.Deck**
-- **CardGames.Card**
+- **Deck**
+- **Card**
 
 `Import-Module` and the `#requires` statement only import the module functions,
 aliases, and variables, as defined by the module. Classes are not imported. The
@@ -142,7 +142,6 @@ aliases, and variables, as defined by the module. Classes are not imported. The
 
 ```powershell
 using module CardGames
-using namespace CardGames
 
 [Deck]$deck = [Deck]::new()
 $deck.Shuffle()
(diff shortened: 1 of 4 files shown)


reference/5.1/Microsoft.PowerShell.Core/About/about_Using.md
```diff
@@ -153,32 +152,25 @@ $deck.Shuffle()
...
-PowerShell classes. The following script creates a new PowerShell class that is
-derived from **DirectoryContext** class.
+This example loads an assembly so that its classes can be used when processing
+data. The following script uses custom settings for serializing a hash table.
 
 ```powershell
-using assembly 'C:\Program Files\PowerShell\7\System.DirectoryServices.dll'
-using namespace System.DirectoryServices.ActiveDirectory
+using assembly 'C:\Program Files\PowerShell\7\Newtonsoft.Json.dll'
+using namespace Newtonsoft.Json
```
michaeltlombardi, line 160: > **Note**: Placed in draft until we address this

I used Newtonsoft here, but I think it might be more useful to instead specify [YamlDotNet][01] instead, since there's no built-in functionality for YAML and coming up with a coherent example for requiring custom JSON serialization eluded me.

[01]: https://github.com/aaubry/YamlDotNet

sdwheeler: Yeah, and Newtonsoft.Json gets loaded by PowerShell. We need an assembly that doesn't ship with PowerShell.

PR #9793, opened 2023-02-03
Update 6.17 remove untrue claim about parameter binding type conversion for boolean

1NF053C:
In order to specify bool with flagging system, you have to supply an argument or else you get an error. This is different behavior than the `switch` type, which allows you to specify the parameter without an arg value without throwing an error. The current state docs indicate that the type conversion for both a `switch` and `bool` are the same, for the case where no arg value is supplied however this is not true in practice. In practice, if a user does not supply an argument value for a `bool`, PowerShell will error, it will not convert that `bool` into a `$true` value.

This PR changes the docs so that the docs no longer imply that passing no argument to a `bool` parameter results in a `$true` value.

```powershell
    Example1 -MySwitch # this is OK, and $MySwitch will be converted to $true
    Example1 -MyBool # this will actually error, it will not convert $MyBool to be $true
    # Section 6.17 bullet 1 in current state docs suggests this second example would convert $MyBool to $true and not throw an error
```

# PR Summary

This PR changes the docs so that the docs no longer imply that passing no argument to a `bool` parameter results in a `$true` value.

## PR Checklist

### DIFF
--- a/reference/docs-conceptual/lang-spec/chapter-06.md
+++ b/reference/docs-conceptual/lang-spec/chapter-06.md
@@ -256,7 +256,7 @@ For information about parameter binding see [§8.14][§8.14].
 When the value of an expression is being bound to a parameter, there are extra conversion
 considerations, as described below:
 
-- If the parameter type is switch ([§8.10.5][§8.10.5]) and the parameter has no
+- If the parameter type is switch ([§4.2.5][§4.2.5], [§8.10.5][§8.10.5]) and the parameter has no
   argument, the value of the parameter in the called command is set to `$true`. If the parameter
   type is other than switch, a parameter having no argument is in error.
 - If the parameter type is switch and the argument value is `$null`, the parameter value is set to
--- a/reference/docs-conceptual/lang-spec/chapter-06.md
+++ b/reference/docs-conceptual/lang-spec/chapter-06.md
@@ -256,9 +256,9 @@ For information about parameter binding see [§8.14][§8.14].
 When the value of an expression is being bound to a parameter, there are extra conversion
 considerations, as described below:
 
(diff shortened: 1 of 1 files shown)


1NF053C:
Greetings,

Thanks to maintainers and community working on PowerShell and PowerShell docs. I'm new to PowerShell and MSFT ecosystem and scratched my head a little bit when I tried understanding this section of the docs. I'm hoping I didn't misinterpret anything and that this edit can make things a little bit easier on the next person.

Please let me know if I need to improve anything about this PR.

Thanks!

reference/docs-conceptual/lang-spec/chapter-06.md
```diff
@@ -256,9 +256,9 @@ For information about parameter binding see [§8.14][§8.14].
 When the value of an expression is being bound to a parameter, there are extra conversion
 considerations, as described below:
 
-- If the parameter type is bool or switch ([§4.2.5][§4.2.5], [§8.10.5][§8.10.5]) and the parameter has no
+- If the parameter type is switch ([§8.10.5][§8.10.5]) and the parameter has no
```
sdwheeler, line 259:
```suggestion
- If the parameter type is switch ([§4.2.5][§4.2.5], [§8.10.5][§8.10.5]) and the parameter has no
```

sdwheeler (approved):
@1NF053C Thanks for catching this. I made one minor change. I restored the link to 4.2.5 because it was relevant to `switch`.

sdwheeler:
@1NF053C I think I know how this confusion was introduced into the doc.

`[switch]` type parameters default to `$false` when not used.

Similarly, in your example, `$MyBool` will be `$false` when the **MyBool** parameter is not used. But that's because any `[bool]` type variable defaults to `$false` unless you assign another value.

PR #9799, opened 2023-02-04
Removes > symbol from content between tags

thebarrettlo:
Minor typo fix


sdwheeler (approved):
@thebarrettlo Thanks for the fix.

PR #9800, opened 2023-02-04
Fixes newlines breaking Markdown table

thebarrettlo:
# PR Summary

Fixes a Markdown table with invalid formatting, resulting in a misrendering of the table.


sdwheeler (approved):
@thebarrettlo Thanks for the fix.

PR #10060, opened 2023-05-08
Fixes #10046 - Update description of LASTEXITCODE

### LINKED ISSUE #10046: Incorrect / incomplete $LASTEXITCODE / exit-code information with respect to script files and -File CLI calls
### Prerequisites

### Links

https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_Automatic_Variables?view=powershell-7.4#lastexitcode

### Summary

The following statements are incorrect:

> When a script is called directly, like ./Test.ps1, or with the [call operator](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_operators?view=powershell-7.4#call-operator-) (&) like & ./Test.ps1, the value of $LASTEXITCODE isn't changed.

`$LASTEXITCODE` _is_ changed:

* _deliberately_, with `exit` statements (with `exit` alone setting `$LASTEXITCODE` to `0`, and `exit $number` setting it to the value of `$number`, if the value is a number of can be coerced to one, otherwise `0`)

* _incidentally_, if the script itself happens to call at least one external program or another script that uses `exit`.

> When a script is called with `powershell.exe` and the `File` parameter, the value of `$LASTEXITCODE` is set to `1` if the script terminated due to a thrown exception and `0` otherwise.

* It is true that a script-terminating (fatal) error sets _exit code_ `1` for the _external caller_, as the exit code of the PowerShell CLI process (`$LASTEXITCODE` is PowerShell session-internal construct, which in the case of a fatal error doesn't apply), and that applies equally to `-Command` calls.

* However, more importantly, a script using an `exit $number` call to exit with an explicit exit code  _is_ honored with `-File`, in that this exit code becomes the PowerShell CLI process' exit code.

  * In the absence of an `exit` call inside a script, `0` is used as the exit code, irrespective of what happens inside the script (short of a _fatal_ error - see above), which notably contrasts with the in-session `$LASTEXITCODE` behavior (see https://github.com/PowerShell/PowerShell/issues/11712 for a discussion).

* By contrast, when using the `-Command` CLI parameter, it is the success status of the _last command executed_ that determines the exit code, as implied by the `$?` variable: `$true` translates to `0`, `$false` to `1`.
  * Thus, the only way to set a _specific_ exit code (other than `1`) is to append an explicit `exit` call to the command string, e.g, `; exit $LASTEXITCODE` (though care needs to be taken not to accidentally surface an _incidental_ `$LASTEXITCODE` value from a script that calls external programs but doesn't use `exit`)

As an aside, it's also worth mentioning `pwsh[.exe]`, the PowerShell _Core_ CLI, and perhaps link to their respective help topics.

### Details

_No response_

### Suggested Fix

_No response_

sdwheeler:
# PR Summary

Update description of LASTEXITCODE

- Fixes [AB#89271](https://dev.azure.com/msft-skilling/cebd7ef5-4282-448b-9701-88c8637581b7/_workitems/edit/89271)
- Fixes #10046

## PR Checklist

### DIFF
--- a/reference/5.1/Microsoft.PowerShell.Core/About/about_Automatic_Variables.md
+++ b/reference/5.1/Microsoft.PowerShell.Core/About/about_Automatic_Variables.md
@@ -1,7 +1,7 @@
 ---
 description: Describes variables that store state information for PowerShell. These variables are created and maintained by PowerShell.
 Locale: en-US
-ms.date: 03/07/2023
+ms.date: 05/08/2023
 no-loc: [Reset, Current, Background, Blink, Bold, Foreground, Formatting, Hidden, Italic, Reset, Reverse, Underline]
 online version: https://learn.microsoft.com/powershell/module/microsoft.powershell.core/about/about_automatic_variables?view=powershell-5.1&WT.mc_id=ps-gethelp
 schema: 2.0.0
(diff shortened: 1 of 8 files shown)


michaeltlombardi (approved):

mklement0:
Again, thanks for the quick turnaround.

There's a typo: `ppowershell.exewsh`

sdwheeler:
> There's a typo: `ppowershell.exewsh`

Ooops! Thanks.

PR #10165, opened 2023-06-11
Add backtick escape

verhovsky:
# PR Summary

It seems that PowerShell accepts escaping single quotes in double quoted strings but this is not documented

```
PS /Users> "``"
`
```

## PR Checklist

### DIFF
--- a/reference/7.3/Microsoft.PowerShell.Core/About/about_Special_Characters.md
+++ b/reference/7.3/Microsoft.PowerShell.Core/About/about_Special_Characters.md
@@ -41,6 +41,7 @@ PowerShell recognizes these escape sequences:
 |  `` `t ``   | Horizontal tab                                  |
 | `` `u{x} `` | Unicode escape sequence (added in PowerShell 6) |
 |  `` `v ``   | Vertical tab                                    |
+| ``` `` ```  | Backtick                                        |
 
 PowerShell also has a special token to mark where you want parsing to stop. All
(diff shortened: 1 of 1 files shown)


michaeltlombardi:
Thanks for this submission, @verhovsky!

In the ["Handling special characters" section of _About Parsing_][01], the documentation notes:

> The backtick character (`` ` ``) can be used to escape any special character in an expression. This is most useful for escaping the argument-mode metacharacters that you want to use as literal characters rather than as a metacharacter. For example, to use the dollar sign (`$`) as a literal in an expandable string:
>
> ```powershell
> "The value of `$ErrorActionPreference is '$ErrorActionPreference'."
> ```
> ```Output
> The value of $ErrorActionPreference is 'Continue'.
> ```

What you've noted in your PR body is a specific instance of the general case for using the backtick character to "escape any special character in an expression."

However, the table you've updated in this PR is describing _escape sequences_, which are a different sort of special case. Escape sequences are ways to "represent characters that aren't part of the standard character set."

Neither the backtick nor dollar sign belong in this table, as both belong to the standard character set. For this reason, I'm closing the PR. Again, thank you for the contribution! Even though we're not accepting this one for technical reasons, we deeply appreciate you taking the time to file the change.

[01]: https://learn.microsoft.com/powershell/module/microsoft.powershell.core/about/about_parsing#handling-special-characters

PR #10342, opened 2023-08-16
Add section to Enum docs for how to import enums from modules

deadlydog:
# PR Summary

Enums work similarly to classes in that the enum type is not imported when using `Import-Module`, and you must use `using module` if you want to reference the enum name in your scripts.

This PR adds a section to the Enum documentation that is similar to [the section in the Class documentation](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_classes?view=powershell-7.3&source=docs#importing-classes-from-a-powershell-module) on how to import enums defined in modules into your scripts.

## PR Checklist

- [x] **Descriptive Title:** This PR's title is a synopsis of the changes it proposes.
- [x] **Summary:** This PR's summary describes the scope and intent of the change.
- [x] **Contributor's Guide:** I have read the [contributors guide][contrib].
- [x] **Style:** This PR adheres to the [style guide][style].

[contrib]: https://learn.microsoft.com/powershell/scripting/community/contributing/overview
[style]: https://learn.microsoft.com/powershell/scripting/community/contributing/powershell-style-guide

PR #10343, opened 2023-08-16
Clarify `using` statement and class/enum definition requirements

deadlydog:
# PR Summary

Classes need to be defined directly in the module's psm1 file in order to be imported into the script using the module, but this wasn't obvious from the previous wording.

This PR adds wording to explicitly mention that classes should be placed in the module's psm1 file.

## PR Checklist

### DIFF
--- a/reference/5.1/Microsoft.PowerShell.Core/About/about_Enum.md
+++ b/reference/5.1/Microsoft.PowerShell.Core/About/about_Enum.md
@@ -1,7 +1,7 @@
 ---
 description: The `enum` statement is used to declare an enumeration. An enumeration is a distinct type that consists of a set of named labels called the enumerator list.
 Locale: en-US
-ms.date: 06/07/2023
+ms.date: 08/17/2023
 online version: https://learn.microsoft.com/powershell/module/microsoft.powershell.core/about/about_enum?view=powershell-5.1&WT.mc_id=ps-gethelp
 schema: 2.0.0
 title: about Enum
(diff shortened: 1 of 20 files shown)


sdwheeler (approved):

deadlydog:
Thanks for incorporating changes from my other PR #10342 into this one. Also, thanks for updating the `about_Using.md` page to include information for enums as well. I had actually written up those changes as well, but was going to open a separate PR after this one was merged in to avoid conflicts. Good to know that it is ok to just put multiple related changes into one larger PR, rather than requiring multiple separate but related PRs 👍

I'm happy with all of the changes you made @sdwheeler and I think it adds a lot of additional clarification that my original changes did not provide. One thing I find that is still missing though is the clarification that when the docs say "the module" they are referring specifically to the .psm1 file. When reading the docs before I equated "the module" to be any files that end up in a published module (e.g. any files in the module's directory that get pulled into the psm1 file). This tripped me up and cost me several hours of troubleshooting, and from the numerous stack overflow questions and open source repos doing things the wrong way, it seems it trips a lot of others up as well. I think it's worth adding a few words or a sentence essentially stating that "the 'root module' is the .psm1 file in a script module".

I'm not sure if we need to open another PR for that change, or if it can be incorporated into this one somehow since it has already been merged?

reference/5.1/Microsoft.PowerShell.Core/About/about_Enum.md
```diff
@@ -327,3 +327,35 @@ to enumeration values that are not valid. Specify one of the following
...
+**Force** parameter. This works for changes to functions in the root module
+only. `Import-Module` doesn't reload any nested modules. Also, there's no way
+to load any updated classes.
+
+To ensure that you're running the latest version, you must start a new session.
+Classes and enumerations defined in PowerShell and imported with a `using`
+statement can't be unloaded.
+
+Another common development practice is to separate your code into different
+files. If you have function in one file that use enumerations defined in
```
deadlydog, line 357: > files. If you have function in one file that use enumerations defined in

to 

> files. If you have a function in one file that use enumerations defined in

deadlydog:
> @michaeltlombardi Did all of the changes.
> 
> 
> 
> > I also thought that the "root module" was a way of referring to the module you were writing
> 
> 
> 
> The root module is defined by the [RootModule](https://learn.microsoft.com/powershell/module/microsoft.powershell.core/about/about_module_manifests#rootmodule) property in the module manifest.

Ah, thanks. I had not seen that Module Manifest page. I think a cross link to the page from the About Using would be helpful then 🙂.

PR #10378, opened 2023-08-30
Dynamic installation script for Raspberry Pi OS

JamesDBartlett3:
- Detect Raspberry Pi OS kernel mode (32-bit or 64-bit)
- Query GitHub Releases API for latest PowerShell release files
- Parse results for the tar.gz file whose architecture matches the detected kernel mode, and store it in the `latest_build` variable
- Use `wget` to download file at `latest_build`
- Parse downloaded tar.gz filename from `latest_build` and extract its contents to `~/powershell`

# PR Summary

## PR Checklist

### DIFF
--- a/reference/docs-conceptual/install/install-raspbian.md
+++ b/reference/docs-conceptual/install/install-raspbian.md
@@ -44,16 +44,23 @@ sudo apt-get update
 sudo apt-get install libssl1.1 libunwind8 -y
 
 ###################################
+
+# Get Raspberry Pi OS kernel mode
+uname_string=$(uname -m | tail -c 3)
+kernel_mode="${uname_string/71/32}"
+echo "Detected kernel mode: ${kernel_mode}-bit"
+
 # Download and extract PowerShell
 
-# Grab the latest tar.gz
-wget https://github.com/PowerShell/PowerShell/releases/download/v7.3.6/powershell-7.3.6-linux-arm32.tar.gz
+# Grab the latest tar.gz for the detected kernel mode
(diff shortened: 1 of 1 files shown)


sdwheeler:
@JamesDBartlett3 Thanks for the contribution. This looks like a great option for your personal toolkit. However, for the documentation, we need to keep the logic as simple as possible.

JamesDBartlett3:
@sdwheeler When I ran the code from the current installation instructions on my Raspberry Pi, it didn't work because my Raspberry Pi is 64-bit. If you're trying to make this "as simple as possible," then it really should automatically detect the current system's kernel mode and install the appropriate version of PowerShell based on that.

sdwheeler:
There are several issues with this.
- Your changes don't work for 32-bit - it should test for `7l` not `71`
- A better way to test for 32/64-bit is to use the command `getconf LONG_BIT`
- We have only ever supported 32-bit for Raspberry Pi
- We no longer officially support Raspberry Pi because we no longer test it - so the docs need to be updated

JamesDBartlett3:
>     * Your changes don't work for 32-bit - it should test for `7l` not `71`
I already fixed that in my fork, but since you closed the PR, that commit isn't showing up here.

>     * A better way to test for 32/64-bit is to use the command `getconf LONG_BIT`
OK, I can do that. Why wasn't this suggested before, instead of just closing the PR?

>     * We have only ever supported 32-bit for Raspberry Pi
Then why is there a 64-bit ARM tarball in the releases page?

>     * We no longer officially support Raspberry Pi because we no longer test it - so the docs need to be updated
If the Raspberry Pi is no longer tested or supported, why does this page still exist at all?

sdwheeler:
The 64-bit ARM tarballs are supported for other OS versions. I am in the process of removing the Raspberry Pi documentation from the list of supported versions. Stay tuned for updates.

PR #10456, opened 2023-09-25
(GH-10451) Extend grouping operator docs

### LINKED ISSUE #10451: Mention in-pipeline uses of `(...)`, the grouping operator
### Prerequisites

### Links

https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_Operators?view=powershell-7.4#grouping-operator--

### Summary

`(...)` is not only necessary to change operation precedence in expressions and in order for commands to participate _in expressions_, it is also useful _in pipelines_, namely in the following scenarios:

* To force _up-front collection_ of a command's output, so that subsequent object-by-object processing cannot interfere with the enumeration the command uses to produce its output.

  * An example is `(Get-ChildItem ...) | Rename-Item -NewName { ... }` - the `(...)` enclosure prevents the `Get-ChildItem` enumeration from being affected by the subsequent renaming.

  * Note: Strictly speaking, this is no longer necessary in PS Core, due to an _implementation detail_, but it applies to WinPS, and similar scenarios with other (potentially custom) cmdlets may exist.

* To force _enumeration_ of a command's output, in case it (unexpectedly) outputs collections _as a whole_, instead of streaming elements one by one.

  * An example - again no longer directly applicable to PS Core, but other scenarios may exist - is to force `ConvertFrom-Json` to output the elements of a JSON array one by one (in WinPS, a _single_ object of type `[object[]]` is output, `(...)` forces its enumeration):

        (ConvertFrom-Json '[1, 2]') | ForEach-Object { ... }

It's worth mentioning these scenarios, given that use of `(...)` in these cases is a concise technique that is a less verbose alternative to capturing command output in an aux. variable in a separate step.

### Details

_No response_

### Suggested Fix

_No response_

michaeltlombardi:
# PR Summary

Prior to this change, the documentation for the grouping operator mentioned using the grouping operator to force enumeration for the pipeline, but not preventing processing the output of a command in a way that affects it.

This change:

- Adds examples for both uses to the Windows PowerShell docs
- Adds a paragraph describing the processing-conflict issue for all versions.
- Resolves #10451
- Fixes #163613

## PR Checklist


sdwheeler (approved):

mklement0:
Thanks for the quick turnaround.

I had missed the "Piping grouped expressions" section before, because the (lengthy) "Grouping assignment statements" comes _before_ it in the enclosing "Grouping operator ( )" section.

I think it would be more easily discoverable if "Piping grouped expressions" came first.
Happy to open a new issue, if necessary.

mklement0:
Please see #10459

sdwheeler:
@microsoft-github-policy-service rerun

PR #10702, opened 2023-12-02
Fixes #10696 - Add info about using environment variable

### LINKED ISSUE #10696: Make it clear that the `POWERSHELL_UPDATECHECK` environment variable must be set *before* PowerShell is launched
### Type of issue

Missing information

### Feedback

As evidenced by the exchange in

*  https://github.com/PowerShell/PowerShell/issues/20742

it isn't obvious that the `POWERSHELL_UPDATECHECK` environment variable must be set _before_ PowerShell is launched in order to be effective, so I suggest clarifying that.

As for providing _guidance_ on how to ensure the above, the following contains all relevant information.
I realize that it goes into way too much detail, because I wrote it up in the context of exploring persistent environment-variable definitions from a cross-platform perspective _in general_, but I hope it provides enough pointers for a succinct update to the topic at hand.

----

* On **Windows**, [`[System.Environment::SetEnvironmentVariable()`](https://learn.microsoft.com/en-US/dotnet/api/System.Environment.SetEnvironmentVariable) is the simplest way to set an environment variable _persistently_; e.g. to set it at the _user_ level:

  * `[Environment]::SetEnvironmentVariable('POWERSHELL_UPDATECHECK', 'Off', 'User')`

  * Replace `'User'` with `'Machine'` to define the variable _system-wide_, i.e. for _all_ users, but note that doing so requires _elevation_ (running with admin privileges).

  * Unfortunately, this API does _not_ work on Unix-like platforms for _persistent_ definitions (`'User'` or `'Machine'`) - such calls are _quietly ignored_. 
 
    * (It does work with `'Process'` as the target, that is for defining an environment variable for the _current process_ - and any child processes it launches - only, but for that you can more easily use PowerShell's convenient `$env:` syntax, e.g. `$env:FOO = 'bar'`)
 
* On **macOS**:

  * There is no system-wide mechanism per se, but _individual shells_ offer _shell-specific_ initialization files.

  * For _system-wide_ environment-variable definitions, the best way to target _multiple_ shells is to use `/etc/profile`, which is read by _most_ POSIX-compatible shells when invoked as _login_ shells (as is the default on macOS), including, indirectly, by PowerShell.

    * Add the following to `/etc/profile` (requires `sudo`): `export POWERSHELL_UPDATECHECK=Off`

    * Since PowerShell, when invoked as a login shell on Unix-like platforms (e.g., with CLI parameter `-l`), _reinvokes itself_ via a `/bin/sh -l` process that it inherits its environment from, placing environment-variable definitions in `/etc/profile` indeed makes them take effect _before_ PowerShell is (re)started.

    * Notably, `zsh` - which is now the default shell on macOS - does _not_ source `/etc/profile` by default. To make it read this file too, add the line `source /etc/profile`  to `/etc/zprofile`

  * For _user-level_ environment-variable definitions you can use `~/.profile`, which is also read by `/bin/sh -l` and therefore takes effect for PowerShell too.

    * Note: `bash` doesn't read this file if a `~/.bash_profile` files is also present, and `zsh` doesn't read it at all; however, both these shells read it when run in POSIX compatibility mode (such as if started from a symlink named `sh` or by explicitly setting `argv[0]` to `sh` or `/bin/sh` - in fact, on macOS it is the latter mechanism that PowerShell uses: instead of `/bin/sh -l`, `/bin/zsh -l` is called with a modified `argv[0]`).

    * Therefore, if you also use these shells directly, use their specific initialization files, i.e., `~/.bash_profile` for `bash` and `~/.zhrc` for `zsh`, but note that these files are by definition read _in-session_, i.e. _after_ startup.
The PowerShell-only analog to such in-session definitions would be to place statements such as `$env:FOO = 'bar'` in a [`$PROFILE`](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_Profiles) file.

    *  If you want to _centralize_ definitions, place them in `~/.profile` and _explicitly `source` (`.`) this file_ from the shell-specific initialization files, using the following POSIX-compliant statement: `[ -f ~/.profile ] && source ~/.profile`
    
* On **Linux** distros (using the same `export` syntax as on macOS):

  * For _system-wide_ environment-variable definitions:

    * Use `/etc/environment`, if present, such as on Ubuntu (requires `sudo`).

    * Absent that, use `/etc/profile`, as on macOS (requires `sudo`), with the _same limitations_.

      * The caveat is that on _headless / non-GUI_ distros, such as Docker containers, you'll have to invoke PowerShell as `pwsh -l` in order for this file to be read, given that on Linux - unlike on macOS - shells are normally _not_ invoked as login shells. However, you can use _per-user_ initialization files instead (see next point).
         
  * For _user-level_ environment-variable definitions you can use `~/.profile` too, as on macOS - again with the same limitations - but you then _must_ invoke PowerShell with `pwsh -l` in order for this file to be read.

    * For general-purpose environment-variable definitions that don't require them to have been made _before_ a given shell is launched, you can again use shell-specific user-level initialization files (including those that are read even by _non-login_ shells) and possibly centralize definitions, as suggested above.

### Page URL

https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_update_notifications?view=powershell-7.3

### Content source URL

https://github.com/MicrosoftDocs/PowerShell-Docs/blob/main/reference/7.3/Microsoft.PowerShell.Core/About/about_Update_Notifications.md

### Author

@sdwheeler

### Document Id

34accac6-39f9-dd49-c540-170e72dc11bc

sdwheeler:
# PR Summary

Add info about using environment variable

- Fixes [AB#187724](https://dev.azure.com/msft-skilling/cebd7ef5-4282-448b-9701-88c8637581b7/_workitems/edit/187724)
- Fixes #10696

## PR Checklist

### DIFF
--- a/reference/7.2/Microsoft.PowerShell.Core/About/about_Update_Notifications.md
+++ b/reference/7.2/Microsoft.PowerShell.Core/About/about_Update_Notifications.md
@@ -47,7 +47,7 @@ You must set this environment variable before PowerShell starts. For more
 information about setting environment variables, see
 [about_Environment_Variables][01].
 
-## Update notification endpoints
+## Notification endpoints
 
 PowerShell uses the following endpoints for determining the latest version
 available for each channel:
--- a/reference/7.2/Microsoft.PowerShell.Core/About/about_Update_Notifications.md
+++ b/reference/7.2/Microsoft.PowerShell.Core/About/about_Update_Notifications.md
@@ -1,7 +1,7 @@
 ---
 description: Notifies users on startup of PowerShell that a new version of PowerShell has been released.
 Locale: en-US
-ms.date: 01/10/2020
+ms.date: 12/02/2023
 online version: https://learn.microsoft.com/powershell/module/microsoft.powershell.core/about/about_update_notifications?view=powershell-7.2&WT.mc_id=ps-gethelp
 schema: 2.0.0
 title: about Update Notifications
@@ -22,17 +22,18 @@ queries an online service to determine if a newer version is available.
 
 > [!NOTE]
 > While the update check happens during the first session in a given 24-hour
-> period, for performance reasons, the notification will only be shown on the
-> start of subsequent sessions. Also for performance reasons, the check will
-> not start until at least 3 seconds after the session begins.
+> period, for performance reasons, PowerShell shows the notification on the
+> start of subsequent sessions. Also for performance reasons, the check for
+> updates starts 3 seconds after the session begins.
 
-By default, PowerShell subscribes to one of two different notification channels
-depending on its version/branch. Supported, Generally Available (GA) versions of
-PowerShell only return notifications for updated GA releases. Preview and
-Release Candidate (RC) releases notify of updates to preview, RC, and GA
-releases.
+By default, PowerShell subscribes to different notification channels depending
+on its version and branch. Generally Available (GA) versions of PowerShell only
+return notifications for updated GA releases. Preview and Release Candidate
+(RC) versions notify of updates to preview, RC, and GA releases.
 
-The update notification behavior can be changed using the
+## Manage notification behavior
+
+You can change the behavior of the update notification by setting the
 `POWERSHELL_UPDATECHECK` environment variable. The following values are
 supported:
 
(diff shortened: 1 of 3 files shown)


mklement0:
🙏

PR #10703, opened 2023-12-03
Fix New-Service -BinaryPathName quotes

shyguyCreate:
# PR Summary

In the documentation for the `New-Service` command, it states that it should be use like the following:

```powershell
New-Service -Name "TestService" -BinaryPathName '"C:\WINDOWS\System32\svchost.exe -k netsvcs"'
```

But the reality is that using this command, the created service will try to find a file with the name `'svchost.exe -k netsvcs'`, giving the message in the Services app `Error 2: The system cannot find the file specified`.

As the documentation says `If the path contains a space, it must be quoted so that it is correctly interpreted.`, but arguments are not part of the path so they should be outside of the double quotes.

## PR Checklist

### DIFF
--- a/reference/5.1/Microsoft.PowerShell.Management/New-Service.md
+++ b/reference/5.1/Microsoft.PowerShell.Management/New-Service.md
@@ -34,7 +34,7 @@ dependencies of the service.
 ### Example 1: Create a service
 
 ```powershell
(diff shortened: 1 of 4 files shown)


sdwheeler (approved):
Thanks for the fix.

shyguyCreate:
@microsoft-github-policy-service terminate

PR #10740, opened 2023-12-17
Fix a few typos

Blake-Madden:
Article mismatches, misspellings, double words


sdwheeler (approved):
@Blake-Madden Thanks again!

PR #10854, opened 2024-02-05
Update install-debian.md

prashanthpulisetti:
updated if source /ect/os-release doesn't work

### DIFF
--- a/reference/docs-conceptual/install/install-debian.md
+++ b/reference/docs-conceptual/install/install-debian.md
@@ -42,6 +42,9 @@ sudo apt-get install -y wget
 # Get the version of Debian
 source /etc/os-release
 
+#If above command doesn't work try:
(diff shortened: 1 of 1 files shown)


sdwheeler:
@prashanthpulisetti Thanks for your contribution. However, it the `source /etc/os-release` command doesn't work then the rest of the script won't. Adding `cat /etc/os-release` doesn't help.

prashanthpulisetti:
Actually I tested today and when I did source /etc/os-release on a Ubuntu
machine, it didn't work, but when I ran cat /etc/os-release, it worked.

Sai Prashanth Pulisetti

sdwheeler:
The `source` command doesn't produce any output. It loads the values in `/etc/os-release` as environment variables so that they can be used in the script. The `cat` command does nothing to the environment.

PR #10877, opened 2024-02-14
Update $OutputEncoding

jborean93:
# PR Summary
Updates the documentation for the $OutputEncoding preference variable. This also updates the examples to correctly reflect what this option actually does.

The current description used is incorrect, `$OutputEncoding` only applies to data being piped into external applications. The command line is always text on all platforms that PowerShell runs on so the encoding never applies and PowerShell uses `[Console]::OutputEncoding` to control how it decodes the output from an external application.

The examples used for the 5.1 docs are slightly different to reflect the different defaults and code changes/output for the 5.1 version. The same fundamental behaviour still applies to both though.

## PR Checklist

### DIFF
--- a/reference/5.1/Microsoft.PowerShell.Core/About/about_Preference_Variables.md
+++ b/reference/5.1/Microsoft.PowerShell.Core/About/about_Preference_Variables.md
@@ -978,12 +978,12 @@ Remove-Variable OFS
 
 ## $OutputEncoding
 
-Determines the character encoding method that PowerShell uses when it sends
-text to other applications.
+Determines the character encoding method that PowerShell uses when piping data
+into native applications.
 
-For example, if an application returns Unicode strings to PowerShell, you might
-need to change the value to **UnicodeEncoding** to send the characters
-correctly.
+> [!NOTE]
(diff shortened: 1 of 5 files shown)


michaeltlombardi:
@jborean93 - Thanks for this PR!

Looking through the updated example, it seems to be shifting from intending to show a practical use towards demonstrating behavior. I'm wondering if there's an alternate example we can replace the (incorrect) existing one with.

jborean93:
> Looking through the updated example, it seems to be shifting from intending to show a practical use towards demonstrating behavior. I'm wondering if there's an alternate example we can replace the (incorrect) existing one with.

That’s the problem I don’t know of any practical examples. I’ve never had to use this variable and while I can see it might be needed I don’t know off any ones that actually exist right now. I even asked in the pwsh Discord and didn’t get any reply, in fact people thought this controlled something else.

In light off that I thought at least showing an actual example where the differences where plain to see was at least better than an example that had no relevance to the option.

michaeltlombardi (approved):

jborean93:
Thanks, if I ever come across a practical example I'll try and remember to submit another PR :)

PR #11021, opened 2024-04-11
Fix a few typos

Blake-Madden:
Repeated words, misspellings, article mismatches, wording errors.

This should be my final pass of everything.


sdwheeler (approved):
@Blake-Madden Thanks for the fixes. This really helps with document quality.

Do you have a tool you use to find these problems or are you finding these problems through manual review?

PR #11032, opened 2024-04-17
Fixes #11017 - Clarify syntax of substitution expressions

### LINKED ISSUE #11017: Description of the syntax of `-replace`'s substitution operand is potentially misleading
### Type of issue

Missing information

### Feedback

Quoting from the  "Substitutions in Regular Expressions" section:

> The `<original>` and `<substitute>` operands are subject to rules of the regular expression engine such as character escaping.

* It is only the `<original>` operand that is a _regex_, where escape sequences such as `\n` or `\s` are recognized.

* By contrast, `<substitute>` is _not_ a regex, and the only form of escaping recognized there is that of escaping `$` characters as `$$` - see https://learn.microsoft.com/en-us/dotnet/standard/base-types/substitutions-in-regular-expressions#substitution-elements-and-replacement-patterns

This distinction is important, because users may be tempted to try something like:

```powershell
@'
one
two
'@ -replace '\n', '\r\n'
```

expecting the `\r\n` escape sequences to be expanded - but they're not (result is verbatim `one\r\ntwo`); ``"`r`n"`` is needed in PowerShell.

It gets tricky when the substitution operand also needs to make references to elements of what was matched by the regex, given that such references are `$`-prefixed (e.g. `$1` to refer to the first capture group) and are therefore subject to up-front string-interpolation by PowerShell inside `"..."` - unless the `$` is escaped as `` `$ ``:

```powershell
'foo-bar' -replace '-(.+)', "`n`$1"  # Note the required ` before $1
```

Alternative approaches, using verbatim substrings:

```powershell
'foo-bar' -replace '-(.+)', ("`n" + '$1')

'foo-bar' -replace '-(.+)', ('{0}$1' -f "`n")
```

While the need to escape `$` in substitution operands is already covered in https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_Comparison_Operators#regular-expressions-substitutions, the _combination_ of an expandable string with verbatim parts is not.

Perhaps the note about non-support of escape sequences in the substitution operand is also worth adding to the latter topic, in the "Warning" box, under "Regular expressions substitutions".

### Page URL

https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_Regular_Expressions?view=powershell-7.4#substitutions-in-regular-expressions

### Content source URL

https://github.com/MicrosoftDocs/PowerShell-Docs/blob/main/reference/7.4/Microsoft.PowerShell.Core/About/about_Regular_Expressions.md

### Author

@sdwheeler

### Document Id

dd3b7f73-ba1e-08b6-45fd-8120305bac72

sdwheeler:
# PR Summary

Clarify syntax of substitution expressions

- Fixes [AB#245895](https://dev.azure.com/msft-skilling/cebd7ef5-4282-448b-9701-88c8637581b7/_workitems/edit/245895)
- Fixes #11017

## PR Checklist

### DIFF
--- a/reference/5.1/Microsoft.PowerShell.Core/About/about_Regular_Expressions.md
+++ b/reference/5.1/Microsoft.PowerShell.Core/About/about_Regular_Expressions.md
@@ -1,7 +1,7 @@
 ---
 description: Describes regular expressions in PowerShell.
 Locale: en-US
-ms.date: 04/10/2023
+ms.date: 04/17/2024
 online version: https://learn.microsoft.com/powershell/module/microsoft.powershell.core/about/about_regular_expressions?view=powershell-5.1&WT.mc_id=ps-gethelp
 schema: 2.0.0
 title: about Regular Expressions
@@ -16,8 +16,7 @@ Describes regular expressions in PowerShell.
 > [!NOTE]
 > This article will show you the syntax and methods for using regular
 > expressions in PowerShell, not all syntax is discussed. For a more complete
-> reference, see the
-> [Regular Expression Language - Quick Reference](/dotnet/standard/base-types/regular-expression-language-quick-reference).
+> reference, see the [Regular Expression Language - Quick Reference][03].
 
 A regular expression is a pattern used to match text. It can be made up of
 literal characters, operators, and other constructs.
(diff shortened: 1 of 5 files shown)


michaeltlombardi (approved):

mklement0:
🙏

PR #11040, opened 2024-04-22
Fixes #11038 - Clarify which env vars need to be set outside of PowerShell

### LINKED ISSUE #11038: about_Telemetry POWERSHELL_TELEMETRY_OPTOUT
### Type of issue

Typo

### Feedback

 POWERSHELL_TELEMETRY_OPTOUT needs to be set BEFORE PowerShell starts, using Windows settings or similar. so having it referenced as `$env:POWERSHELL_TELEMETRY_OPTOUT` is misleading, as that is PowerShell syntax. if you set the variable using PowerShell itself it will be ignored, as the check happens on startup

### Page URL

https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_telemetry?view=powershell-7.4

### Content source URL

https://github.com/MicrosoftDocs/PowerShell-Docs/blob/main/reference/7.4/Microsoft.PowerShell.Core/About/about_Telemetry.md

### Author

@sdwheeler

### Document Id

7342b95a-d041-0cf9-17f6-816e79654525

### LINKED ISSUE #11039: Clarify that most of the environment variables that PowerShell respects must be set BEFORE starting PowerShell
### Type of issue

Missing information

### Feedback

Note: This issue is a generalization of:

* ... what #10696 asked for (already implemented) with respect to `about_Update_Notifications`

* ... what #11038 now asks for with respect to `about_Telemetry`

----

With few exceptions (see below), well-known environment variables respected by PowerShell (both PowerShell-specific ones and system-wide / cross-application ones) must be set BEFORE launching PowerShell in order to take effect.

* _Ad hoc_, this requires the process that calls PowerShell to define such variables, using whatever mechanism is appropriate for them (e.g., `set POWERSHELL_UPDATECHECK=Off` from `cmd.exe` or `export POWERSHELL_UPDATECHECK=Off` from Bash).

* _Persistently_, a platform-specific mechanism must be used, as previously detailed in https://github.com/MicrosoftDocs/PowerShell-Docs/issues/10696#issue-2019714367

* If you accidentally set these variables _in-process_, they are only respected by PowerShell _child_ processes, if any.

As the linked issues attest, explicitly stating this requirement is called for.

From what I can tell, the only _exceptions_ are (that is, in-process changes to these env. vars. take effect immediately, also for the current session (process)):

* Among PowerShell-specific env. vars.:

  * `$env:PSModulePath`
  * `$env:PSExecutionPolicyPreference` (while setting this variable _directly_ rather than via `Set-ExecutionPolicy` isn't documented as such, it does work).
  * Note that even though `$env:PSDisableModuleAnalysisCacheCleanup` is currently documented as falling into this category, setting the variable in-process cannot prevent a potential cleanup at session _startup_ time: The variable value is first queried  _at startup_, at which point a cleanup check is scheduled with a 10-second delay if the var. isn't defined, and when the cleanup check happens, the variable is _not_ checked again. That is, in-process setting of the variable cannot prevent an _initial_ cleanup (if necessary), though it can prevent _subsequent_ cleanups, if any.

* Among other well-known env. vars. respected by PowerShell:
  * `$env:PATH`
  * On Windows only, also `$env:PATHEXT`

### Page URL

https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_Environment_Variables?view=powershell-7.5#powershell-environment-variables

### Content source URL

https://github.com/MicrosoftDocs/PowerShell-Docs/blob/main/reference/7.5/Microsoft.PowerShell.Core/About/about_Environment_Variables.md

### Author

@sdwheeler

### Document Id

dfc13aff-02f5-c1ea-d736-68a55241ab52

sdwheeler:
# PR Summary

Clarify which env vars need to be set outside of PowerShell

- Fixes [AB#247899](https://dev.azure.com/msft-skilling/cebd7ef5-4282-448b-9701-88c8637581b7/_workitems/edit/247899)
- Fixes #11038
- Fixes #11039

## PR Checklist

### DIFF
--- a/reference/7.2/Microsoft.PowerShell.Core/About/about_Environment_Variables.md
+++ b/reference/7.2/Microsoft.PowerShell.Core/About/about_Environment_Variables.md
@@ -1,7 +1,7 @@
 ---
 description: Describes how to access and manage environment variables in PowerShell.
 Locale: en-US
-ms.date: 02/05/2024
+ms.date: 04/22/2024
 online version: https://learn.microsoft.com/powershell/module/microsoft.powershell.core/about/about_environment_variables?view=powershell-7.2&WT.mc_id=ps-gethelp
 schema: 2.0.0
 title: about Environment Variables
@@ -70,7 +70,7 @@ environment variables.
 - The Environment provider and Item cmdlets
 - The .NET **System.Environment** class
 
-## Using the variable syntax
+## Use the variable syntax
 
 You can display and change the values of environment variables with the
 following syntax:
@@ -157,7 +157,7 @@ Length Property   int Length {get;}
 
 For more information about variables in PowerShell, see [about_Variables][11].
 
-## Using the Environment provider and Item cmdlets
+## Use the Environment provider and Item cmdlets
 
 PowerShell's **Environment** provider gives you an interface for interacting
 with environment variables in a format that resembles a file system drive. It
@@ -210,10 +210,10 @@ Get-ChildItem Env:
 For more information on using the **Environment** provider to manage
 environment variables, see [about_Environment_Provider][04].
 
-## Using the System.Environment methods
+## Use the System.Environment methods
 
-The **System.Environment** class provides the **GetEnvironmentVariable** and
-**SetEnvironmentVariable** methods to get and modify environment variables.
+The **System.Environment** class provides the `GetEnvironmentVariable()` and
+`SetEnvironmentVariable()` methods to get and modify environment variables.
 
 The following example creates a new environment variable, `Foo`, with a value
 of `Bar` and then returns its value.
@@ -227,7 +227,7 @@ of `Bar` and then returns its value.
 Bar
 ```
 
-You can remove an environment variable with the **SetEnvironmentVariable**
+You can remove an environment variable with the `SetEnvironmentVariable()`
 method by specifying an empty string for the variable's value. For example,
 to remove the `Foo` environment variable:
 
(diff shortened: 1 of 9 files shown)


michaeltlombardi (approved):

reference/7.2/Microsoft.PowerShell.Core/About/about_Environment_Variables.md
```diff
@@ -359,7 +378,11 @@ The environment variables that store preferences include:
   - Use the `Set-ExecutionPolicy` cmdlet. Use the **Scope** parameter with
     a value of `Process`.
 
-    For more information, see [about_Execution_Policies][05].
+  - Manually set the environment variable. Changing the value of this variable
+    changes the execution policy current process.
```
mklement0, line 382: Thanks for the quick turnaround; I think "changes the execution policy current process." should be "changes the execution policy for the current process."

reference/7.2/Microsoft.PowerShell.Core/About/about_Environment_Variables.md
```diff
@@ -308,6 +311,16 @@ System Control Panel:
...
 1. Make your changes.
 
+## Create persistent environment variables on non-Windows platforms
+
+Linux and macOS have configuration files and scripts that the operating system
+uses to set environment variables before starting an application. For example
+on Linux, you can add environment variables to the `/etc/environment` file or
+create a script that sets environment variables and put it in the
+`/etc/profile.d` folder. On macOS, you can add environment variables to the
+`/etc/zshrc` file. For more information, see the documentation for your
```
mklement0, line 321: Note that `/etc/zshrc` is the appropriate file to use only if `zsh` is a user's default shell.

If _PowerShell_ is the default shell, `/etc/profile` must be used, because PowerShell, when invoked with `-l` (as happens implicitly on macOS for the default shell), it reinvokes itself via `zsh -l` with the latter in POSIX compatibility mode, where only `/etc/profile` is sourced.

sdwheeler: @mklement0, can you point to documentation for `zsh` that states that. The documentation I found indicated that `/etc/zshrc` would be invoked by `zsh -l`.

mklement0: As also noted in the linked issue, the `/etc/profile` / `/etc/profile.d` approaches on non-GUI Linux distros - e.g. in Docker images - require explicit invocation via `pwsh -l` to take effect.

PR #11077, opened 2024-05-07
Adds reference docs links, formatting and refactoring test performance test

santisq:
# PR Summary

This PR adds many reference links to PowerShell and .NET docs. I'm not sure if the links are correct, please let me know if they should be fixed, I've basically removed `https://learn.microsoft.com/en-us` from them and made them relative hope that's right. 

There are also many formatting changes and also updated the performance tests for __Avoid repeated calls to a function__ to have the same format as the previous tests.

As aside, I feel that __Avoid Write-Host__ should be removed or rewritten. Haven't removed it in this PR but if you feel the same I can open an issue / make a new PR for that. After discussing it in the pwsh Discord the conclusion was that the recommendation is misleading. The usage of static methods from `System.Console` can be confusing specially for new folks and shouldn't be part of this doc, and the recommendation to use `Write-Output` instead of `Write-Host` is also incorrect, first, the explicit usage of `Write-Output` is not needed to begin with and is only useful when preventing pipeline enumeration with `-NoEnumerate` or for unrolling a collection i.e.: `, (0..10) | Write-Output | ForEach-Object { "[$_]" }` vs `, (0..10) | ForEach-Object { "[$_]" }` and second, the purpose of `Write-Host` is to send information directly to the console and prevent it's output to be captured by default. So if the intent is to offer a better cmdlet for this purpose, a better recommendation would be to use `Write-Verbose` instead of `Write-Host` ?

## PR Checklist

### DIFF
--- a/reference/docs-conceptual/dev-cross-plat/performance/script-authoring-considerations.md
+++ b/reference/docs-conceptual/dev-cross-plat/performance/script-authoring-considerations.md
@@ -1,6 +1,6 @@
 ---
 description: Scripting for Performance in PowerShell
-ms.date: 06/23/2023
+ms.date: 05/07/2024
 title: PowerShell scripting performance considerations
 ---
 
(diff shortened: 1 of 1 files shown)


sdwheeler:
I did some minor editing to conform to our style guidelines and updated the notes about `Write-Host`.

PR #11559, opened 2024-12-04
Fix powershell-lts verification command

bharathalleni:
# PR Summary

Update the installation verification command for powershell-lts on macOS, replacing `pwsh` with the correct `pwsh-lts`.


sdwheeler (approved):
Thanks for the update.

PR #11568, opened 2024-12-07
Fixes #10688 - Add about_Type_Conversion article

### LINKED ISSUE #10688: Create an about_Type_Conversions conceptual help topic that describes PowerShell's unusually flexible and automatic type conversions
### Prerequisites

- [X] **Descriptive Title:** Write the title for this issue as a short synopsis. If possible, provide context. For example, "Document new `Get-Foo` cmdlet" instead of "New cmdlet."

### PowerShell Version

5.1, 7.2, 7.3, 7.4

### Summary

PowerShell is unusually flexible with respect to type conversions, both _explicit_ and _implicit_ ones; while the latter ones are usually _helpful_, there are pitfalls, especially for users coming from languages with stricter type handling.

It would be helpful to systematically describe both how explicit and especially implicit type conversions (coercions) work in PowerShell.

Below is what I have gleaned from my own experience and experiments.

### Details

* PowerShell is unusually flexible with respect to type conversions, both _explicit_ and _implicit_ ones; while the latter ones are usually helpful once understood, there are pitfalls, especially for users coming from languages with stricter type handling.

  * PowerShell not only supports many more _explicit_ type conversions than C#, for instance, using _casts_, but also _applies them automatically (implicitly)_ in the absence of casts, if the context requires it; specifically, these contexts are:

    * Binding (passing) a value to a cmdlet / function / script / .NET method _parameter_ declared with a _specific type_ (i.e. _not_ declared without a type or, equivalently, as `[object]`; see further below).

    * A conceptually and technically closely related context is when assigning values to _type-constrained variables_ (also see further below).
    
    * In the context of using _operators_ (see next point).

    * In _Boolean_ contexts, such as in `if`-statement conditionals, and PowerShell notably supports coercing _any_ value to `[bool]` - see [about_Booleans](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_Booleans).

   * As for conversions PowerShell supports _in addition to_ C#:

     * _Numeric types_ can be freely converted to one another (assuming the target type is wide enough), e.g. `[byte] 42.1` (`[double]` coerced to `[byte]` `42`)

     * A value of _any_ type can be coerced to:

       * `[bool]`, i.e `$true` or `$false`, as noted; e.g. `[bool] 42` is `$true`

       * `[string]`, including _arrays_, with the exceptions during parameter-binding noted below; e.g. `[string] @(1, 2, 3)` is `'1 2 3'` by default (the separator - a space by default - can be controlled with the rarely used `$OFS` preference variable)
         *  Note that to- and from string conversion is usually culture-_invariant_, i.e. based on the [invariant culture](https://learn.microsoft.com/en-US/dotnet/api/System.Globalization.CultureInfo.InvariantCulture), which is based on, but not identical to, the US-English culture, and notably uses `.` as the decimal mark and US-style month-first dates by default. There is an unfortunate exceptions, however: _binary cmdlets_ accidentally perform culture-_sensitive_ conversion during parameter binding, and this behavior cannot be corrected without breaking backward compatibility: see https://github.com/PowerShell/PowerShell/issues/6989

     * Single-character `[string]` instances can be converted to and from `[char]` (note that PowerShell has no `[char]` _literals_).
     
     * The values of [`System.Enum`](https://learn.microsoft.com/en-US/dotnet/api/System.Enum)-derived types can be converted to and from `[string]` instances:
       * E.g., `[System.PlatformId] 'Unix'` is the same as `[System.PlatformId]::Unix`
       * Even _flag_-based enums are handled correctly via `,` separated values inside a string or even string _arrays_; e.g., both  `[System.Reflection.TypeAttributes] 'Public, Abstract'` and `[System.Reflection.TypeAttributes] ('Public', 'Abstract')` are equivalent to `[System.Reflection.TypeAttributes]::Public -bor [System.Reflection.TypeAttributes]::Abstract`.

     * A single value (non-array) can be converted to an instance of a type if that type has a (public) single-parameter constructor of the same type (or a type that the value can be coerced to); e.g. `[regex] 'a|b'` is the same as `[regex]::new('a|b')`

     * A single _string_ value can converted to an instance of a type if the type implements a static `::Parse()` method; e.g., `[bigint] '42'` is the same as `[bigint]::Parse('42', [cultureinfo]::InvariantCulture)` - `[cultureinfo]::InvariantCulture` binds to an `IFormatProvider`-typed parameter - if available - and is used to ensure culture-invariant behavior.

     * _Custom_ conversions can be defined:

       * In the context of _parameter binding_, by decorating parameters with [`System.Management.Automation.ArgumentTransformationAttribute`](https://docs.microsoft.com/en-US/dotnet/api/System.Management.Automation.ArgumentTransformationAttribute)-derived attributes.

       * In the context of _ETS type definitions_ (via [`Update-TypeData`](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/update-typedata) `-TypeConverter` or [`.Types.ps1xml` files](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_Types.ps1xml)) or in _types in compiled code decorated with a [`TypeConverterAttribute`](https://learn.microsoft.com/en-US/dotnet/api/System.ComponentModel.TypeConverterAttribute) attribute_, via classes derived from [`TypeConverter`](https://learn.microsoft.com/en-US/dotnet/api/System.ComponentModel.TypeConverter) or [`PSTypeConverter`](https://learn.microsoft.com/en-US/dotnet/api/System.Management.Automation.PSTypeConverter).

     * For the complete story, consult [the source code]( https://github.com/PowerShell/PowerShell/blob/ac84c35820eeaf46db2d37f8b91ed4b5683c8050/src/System.Management.Automation/engine/LanguagePrimitives.cs#L1608)

* PowerShell _operators_ fundamentally do _not_ guarantee that the result of an expression is of the same type as its operands.
  * `1 / 2` (`[int]` / `[int]` -> `[double`) and `'10' - '9'` (`[string] - [string]` -> `[int]`) are two examples.
  * See the next section for details.

* PowerShell _variables_ are by default _not type-constrained_; that is, you can create a variable with an instance of one type, and later assign values of any other type.
  * To _type-constrain_ variables, place a "cast" (type literal) to the LEFT of the variable name in an assignment (e.g. `[int] $foo = 42`)
  * In addition to immediately converting an assigned value to the specified type, later assignments then enforce that a new value either already _is_ of that type or _can be converted to it_; e.g, `$foo = '43'` works too, because PowerShell happily converts a _string_ that can be _parsed as_ an integer to one.

* Number _literals_ are implicitly typed by default (e.g. `42` is of type `[int]` and `1.2` is of type `[double]`), but can be explicitly typed with a _type-specifier suffix_ (e.g., `42L` or `42l` are of type `[long]` aka `[System.Int64]`) - see [about_Numeric_Literals](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_Numeric_Literals).

* Implicit type conversions also happen during _parameter binding_ when calling PowerShell cmdlets, functions, and scripts, as well as .NET methods; e.g.:

     * `& { param([Int] $Integer) $Integer } -Integer '  -10 '` - the `[string]` instance `'  -10 '`is automatically converted to `[int]` `-10`.
     * `[datetime]::FromFileTime('0')` is the same as `[datetime]::FromFileTime(0)`, i.e. string `'0'` is converted to `[int]` `0`
     * _Caveats_: 
       * With .NET methods in particular, it is better to pass the exact type expected, if needed with a _cast_, as there can be ambiguity otherwise, with PowerShell potentially selecting the _wrong_ overload from among the matching ones. In the worst-case scenario, the introduction of additional method overloads in _future .NET versions_ can break existing code - see https://github.com/PowerShell/PowerShell/issues/11720#issuecomment-579866445 for an example.
       * The automatic coercion of _arrays_ to `[string]`-typed parameters - which is usually undesired - can be treacherous:
         * `& { param([string] $String) $String } -String 1, 2` quietly accepts the array argument and _stringifies it_, i.e. passes the equivalent of `"$(1, 2)"`, which is `"1 2"` by default.
         * You can avoid this obscure behavior by making your script  or function an [_advanced_](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_Functions_Advanced) one:
           * `& { [CmdletBinding()] param([string] $String) $String } -String 1, 2` _fails_, because it refuses to bind an array to a (non-array) `[string]` parameter.
         * Unfortunately, you can _not_ avoid this behavior in .NET method calls; the same behavior as in non-advanced function applies; e.g.:
           * `(Get-Date).ToString(@(1, 2))` quietly passes `"1 2"` to the `string`-typed `format` parameter.
---

#### Type coercions (conversions) performed by PowerShell's operators:

* In _numeric_ operations, even if both operands are of the _same_ numeric type, the _result_ may be a different type, due to _automatic, on-demand type-widening_; namely:

  * Widening to `[double]` to support  fractional results in integer division:
    *  E.g. `3 / 2` , despite having two `[int]` operands, yields `1.5`, i.e, a `[double]`
    * Conversely, to get true integer division, use `[int] [Math]::Truncate(3 / 2)` or `[Math]::DivRem(3, 2)[0]` (`[Math]::DivRem(3, 2, [ref] $null)` in Windows PowerShell).

  * Widening to `[double]` to support results that would result in _overflow_ with the operand type.
    * E.g, `[int]::MaxValue + 1` returns `2147483648` - as a `[double]` - instead of overflowing.
    * Note: It is unfortunate that `[double]` is invariably used when this widening occurs, even though the next larger _integer_ type (`[long]` aka `[System.Int64]` in this case). would suffice and be preferable. Sadly, it only works that way in number _literals_,  where, for instance, `2147483647` (the value of `[int]::MaxValue`) is an `[int]`, where as `2147483648` (i.e. `[int]::MaxValue + 1`), implicitly becomes a `[long]`

* In operations where an _implicit type conversion is required_ in order for the operation to succeed:

  * _Typically_, it is the *LHS* operand of PowerShell operators that determines the data type used in the operation and converts (coerces) the RHS operand to the required type; e.g.:
    * `10 - '  +9'` yields `[int]` `1`, because the `[string]`-typed RHS was implicitly converted to `[int]` 
    * `10 -eq '   +10    '` yields `$true` for the same reason, and the same goes even for  `10 -eq '0xa'`

  * There are _exceptions_, however:
     * _Arithmetic_ operators (`+`, `-`, `*`, `/`) with _non-numeric_ operands:
       * `-` and `/` convert *both* operands from _strings_ to _numbers_ on demand; e.g., `' 10' - '2'` yields `[int] 8`; 
       * By contrast, this does _not_ happen with `+` and `*`, which have _string-specific semantics_ (concatenation and replication).  
       * Using `[bool]` values with arithmetic operators causes them to be coerced to `[int]`, with `$true` becoming `1`, and `$false`, `0`
         * Curiously, as of v7.4.0 this doesn't work with `*` - see https://github.com/PowerShell/PowerShell/issues/20816
          * E.g., `$false - $true` is `-1`, because it is the equivalent of `0 - 1`.
          * The implicit to-integer coercion can also be used as a shortcut in _index_ expressions, such as in the following (imperfect) emulation of a ternary conditional: `('even', 'uneven')[1 -eq [datetime]::Now.Second % 2]`; that is, the Boolean result of the index expression was coerced to `0` or `1` to select the array element of interest.

       * For other LHS types, arithmetic operators only succeed if a given type custom-defines these operators via [operator overloading](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/statements-expressions-operators/overloadable-operators).

     * _Polymorphic comparison_ operators such as `-eq`, `-lt`, `-gt`, ... (as distinct from _string-only_ operators such as `-match` and `-like`):

         * For _non-strings_ and _non-primitive_ types, the behavior depends on whether the LHS type implements _interfaces_ such as `IEquatable` and `IComparable`.

       * The _collection_-based comparison operators, namely `-in` and `-contains` (and their negated variants), perform _per-element_ `-eq` comparisons until a match is found, and it is each _individual element_ of the _collection-valued operand_ that drives any coercion; e.g.:
         *  `$true -in 'true', 'false'` and `'true', 'false' -contains $true` are both`$true`, because `'true' -eq $true` yields `$true` (the equivalent of `'true' -eq [string] $true`)
         * `'1/1/70' -in 'one', [datetime]::UnixEpoch, 'two'` is `$true`, because `[datetime]::UnixEpoch -eq '1/1/70'` is `$true` (the equivalent of `[datetime]::UnixEpoch -eq [datetime] '1/1/70'`)

### Proposed Content Type

About Topic

### Proposed Title

about_Type_Conversions

### Related Articles

https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_Operators

sdwheeler:
# PR Summary

Add about_Type_Conversion article

- Fixes [AB#351599](https://dev.azure.com/msft-skilling/cebd7ef5-4282-448b-9701-88c8637581b7/_workitems/edit/351599)
- Fixes #10688

## PR Checklist

### DIFF
--- a/reference/7.5/Microsoft.PowerShell.Core/About/about_Type_Conversion.md
+++ b/reference/7.5/Microsoft.PowerShell.Core/About/about_Type_Conversion.md
@@ -1,0 +1,497 @@
+---
+description: Describes how and when PowerShell performs type conversions.
+Locale: en-US
+ms.date: 12/05/2024
+online version: https://learn.microsoft.com/powershell/module/microsoft.powershell.core/about/about_type_conversion?view=powershell-7.5&WT.mc_id=ps-gethelp
+schema: 2.0.0
+title: about_Type_Conversion
+---
+# about_Type_Conversion
+
+## Short description
+
+PowerShell has a flexible type system that makes it easier to use. However, you
+must understand how it works to avoid unexpected results.
+
+## Long description
+
+PowerShell has a flexible type system that makes it easier to use. However, you
+must understand how it works to avoid unexpected results.
+
+By default, PowerShell _variables_ aren't _type-constrained_. You can create a
+variable containing an instance of one type and later assign values of any
+other type. Also, PowerShell automatically converts values to other types, both
+explicitly and implicitly. While implicit type conversion can be helpful, there
+are pitfalls, especially for users more familiar with languages have stricter
+type handling.
+
+## Type-constrained variables and explicit types conversion
+
+To type-constrain a variable, place a type literal to the left of the
+variable name in an assignment. For example:
+
+```powershell
+[int]$foo = 42
+```
+
+You can use type casting to explicitly convert a value to a specific type. For
+example:
+
+```powershell
+PS> $var = [int]'43'
+PS> $var.GetType().Name
+Int32
+```
+
+Type casting ensures that only values of the specified type can be assigned to
+the variable. If you try to assign a value of a different type that can be
+converted to the constrained type, PowerShell performs an implicit conversion.
+For more information, see the [Implicit type conversion][03] section of this
+article.
+
+### Numeric type conversion
+
+Numeric types can be freely converted to one another as long as the target type
+is wide enough. For example:
+
+```powershell
+PS> (42.1).GetType().Name
+Double
+PS> $byte = [byte] 42.1
+PS> $byte
+42
+PS> $byte.GetType().Name
+Byte
+```
+
+The value `42.1` is a **Double**. When you convert it to a **Byte**, the value
+is truncated to an integer `42`, which is small enough to fit into a **Byte**.
+
+### Boolean type conversion
+
+A value of _any_ type can be coerced to a **Boolean**.
+
+- For numeric types, `0` converts to `$false` and any other value converts to
+  `$true`.
+
+  ```powershell
+  PS> [boolean]0
+  False
+  PS> [boolean]0.0
+  False
+  PS> [boolean]-1
+  True
+  PS> [boolean]1
+  True
+  PS> [boolean]42.1
+  True
+  ```
+
+- For other types, null values, empty strings, and empty arrays are converted
+  to `$false`. Other values, including empty hashtables convert to `$true`.
+
+  ```powershell
+  PS> [boolean]''
+  False
+  PS> [boolean]@()
+  False
+  PS> [boolean]'Hello'
+  True
+  PS> [boolean]@(1,'Hello')
+  True
+  PS> [boolean]@{}
+  True
+  ```
+
+### String type conversion
+
+A value of _any_ type can be coerced to a **String**. The default conversion is
+to call the `ToString()` method on the object.
+
+Arrays are converted to strings. Each element in the array is converted to a
+string, individually and joined to the resulting string. By default, the
+converted values are separated by a space. The separator can be changed by
+setting the `$OFS` preference variable.
+
+```powershell
+PS> [string] @(1, 2, 3)
+1 2 3
+```
+
+For more information about `$OFS`, see [about_Preference_Variables][09].
+
+A single string value can be converted to an instance of a type if the type
+implements a static `Parse()` method. For example, `[bigint]'42'` is the same
+as `[bigint]::Parse('42', [cultureinfo]::InvariantCulture)`. The optional
+`[cultureinfo]::InvariantCulture` value binds to an `IFormatProvider` type
+parameter of the method. It ensures culture-invariant behavior of the
+conversion. Not all implementations of `Parse()` methods have this parameter.
+
+> [!NOTE]
+> Conversion to and from strings is usually performed using the
+> [invariant culture][16]. Invariant culture is based on, but not identical to,
+> the US-English culture. Notably, it uses the period (`.`) as the decimal
+> mark and US-style month-first dates by default. However, _binary cmdlets_
+> perform culture-sensitive conversion during parameter binding.
+
+### Enum type conversion
+
+PowerShell can convert [Enum][15] types to and from **String** instances. For
+example, the typecast string `[System.PlatformId]'Unix'` is the same as the
+enum value `[System.PlatformId]::Unix`. PowerShell also handles flag-based
+enums correctly for comma-separated values inside a string or as an array of
+strings. Consider the following examples:
+
+```powershell
+[System.Reflection.TypeAttributes]'Public, Abstract'
+[System.Reflection.TypeAttributes]('Public', 'Abstract')
+```
+
+These examples are equivalent to the enum expression:
+
+```powershell
+[System.Reflection.TypeAttributes]::Public -bor
+    [System.Reflection.TypeAttributes]::Abstract
+```
+
+### Other type conversions
+
+A single value (non-array) can be converted to an instance of a type if:
+
+- That type has a (public) single-parameter constructor
+- And the value is same type or can be coerced to the type of the parameter
+
+For example, the following two lines are equivalent:
+
+```powershell
+[regex]'a|b'
+[regex]::new('a|b')`
+```
+
+## Implicit types
+
+PowerShell also assigns types to literal values automatically.
+
+Numeric literals are implicitly typed by default. Numbers are typed based on
+their size. For example, `42` is small enough to be stored as an **Int32** type
+and `1.2` is stored as a **Double**. Integers larger than `[Int32]::MaxValue`
+are stored as **Int64**. While `42` can be stored as a **Byte** and `1.2` can
+be stored as a **Single** type, the implicit typing uses **Int32** and
+**Double** respectively. For more information, see
+[about_Numeric_Literals][07].
+
+Literal strings are implicitly typed as **String**. Single-character **String**
+instances can be converted to and from the **Char** type. However, PowerShell
+has no literal **Char** type.
+
+## Implicit type conversion
+
+In certain contexts, PowerShell can implicitly convert values to other types.
+These contexts include:
+
+- Parameter binding
+- Type-constrained variables
+- Expressions using operators
+- Boolean contexts - PowerShell converts the conditional expressions of `if`,
+  `while`, `do`, or `switch` statements to **Boolean** values, as previously
+  described. For more information, see [about_Booleans][04].
+- Extended Type System (ETS) type definitions - Type conversions can be defined
+  in several ways:
+  - Using the **TypeConverter** parameter of [Update-TypeData][12]
+  - In [Types.ps1xml][10] files
+  - In compiled code for types decorated with a [TypeConverterAttribute][14]
+    attribute
+  - Via classes derived from [TypeConverter][13] or [PSTypeConverter][18]
+
+### Parameter binding conversions
+
+PowerShell attempts to convert values passed to parameters to match the
+parameter type. Type conversion of parameter values occurs in cmdlets,
+functions, scripts, scriptblocks, or .NET method where the parameter is
+declared with a specific type. Declaring a parameter with the type `[object]`
+or not defining a specific type allows any value type to be passed to a
+parameter. Parameters can also have custom conversions defined by decorating
+parameters with [ArgumentTransformationAttribute][17] attribute.
+
+For more information, see [about_Parameter_Binding][08].
+
+### Best practices for parameter binding
+
+For .NET methods, it's better to pass the exact type expected using a type
+casts where needed. Without exact types, PowerShell can select the wrong method
+overload. And, new method overloads added in future versions of .NET can break
+existing code. For an extreme example of this problem, see this
+[Stack Overflow question][11].
+
+If you pass an array to a `[string]` typed parameter, PowerShell might convert
+the array to a string as previously described. Consider the following basic
+function:
+
+```powershell
+function Test-String {
+    param([string] $String)
+    $String
+}
+
+Test-String -String 1, 2
+```
+
+This function outputs `1 2` because the array is converted to a string. To
+avoid this behavior, create an [advanced function][06] by adding the
+`[CmdletBinding()]` attribute.
+
+```powershell
+function Test-String {
+    [CmdletBinding()]
+    param([string] $String)
+    $String
+}
+
+Test-String -String 1, 2
+```
+
+For advanced functions, PowerShell refuses to bind the array to a non-array
+type. When you pass an array, PowerShell returns the following error message:
+
+```Output
+Test-String:
+Line |
+   7 |  Test-String -String 1, 2
+     |                      ~~~~
+     | Cannot process argument transformation on parameter 'String'. Cannot
+     | convert value to type System.String.
+```
+
+Unfortunately, you can't avoid this behavior for .NET method calls.
+
+```powershell
+PS> (Get-Date).ToString(@(1, 2))
+1 2
+```
+
+PowerShell converts the array to the string `"1 2"`, which is passed to the
+**Format** parameter of the `ToString()` method.
+
+The following example shows another instance of the array conversion problem.
+
+```powershell
+PS> $bytes = [byte[]] @(1..16)
+PS> $guid = New-Object System.Guid($bytes)
+New-Object: Cannot find an overload for "Guid" and the argument count: "16".
+```
+
+PowerShell treats the `$bytes` array as a list of individual parameters. Even
+though `$bytes` is an array of bytes and `System.Guid` has a `Guid(Byte[])`
+constructor.
+
+This common code pattern is instance of _pseudo method syntax_, which doesn't
+always work as intended. This syntax translates to:
+
+```powershell
+[byte[]] $bytes = 1..16
+New-Object -TypeName System.Guid -ArgumentList $bytes # !! BROKEN
+```
+
+Given that the type of **ArgumentList** is `[object[]]`, a single argument
+that happens to be an array (of any type) binds to it _element by element_. The
+workaround is to wrap `$bytes` in an outer array so that PowerShell looks for
+constructor with parameters that match the contents of the outer array.
+
+```powershell
+PS> $guid = New-Object System.Guid(@(, $bytes))
+PS> $guid
+
+Guid
+----
+04030201-0605-0807-090a-0b0c0d0e0f10
+```
+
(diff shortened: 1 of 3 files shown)


reference/7.5/Microsoft.PowerShell.Core/About/about_Type_Conversion.md
```diff
@@ -0,0 +1,497 @@
...
+## Long description
+
+PowerShell has a flexible type system that makes it easier to use. However, you
+must understand how it works to avoid unexpected results.
+
+By default, PowerShell _variables_ aren't _type-constrained_. You can create a
+variable containing an instance of one type and later assign values of any
+other type. Also, PowerShell automatically converts values to other types, both
+explicitly and implicitly. While implicit type conversion can be helpful, there
+are pitfalls, especially for users more familiar with languages have stricter
```
mikefrobbins, line 25:
```suggestion
are pitfalls, especially for users more familiar with languages that have stricter
```

reference/7.5/Microsoft.PowerShell.Core/About/about_Type_Conversion.md
```diff
@@ -0,0 +1,497 @@
...
+
+```powershell
+PS> $var = [int]'43'
+PS> $var.GetType().Name
+Int32
+```
+
+Type casting ensures that only values of the specified type can be assigned to
+the variable. If you try to assign a value of a different type that can be
+converted to the constrained type, PowerShell performs an implicit conversion.
```
mikefrobbins, line 48:
```suggestion
the variable. PowerShell performs an implicit conversion if you try to assign a value of a different type that can be converted to the constrained type.
```

reference/7.5/Microsoft.PowerShell.Core/About/about_Type_Conversion.md
```diff
@@ -0,0 +1,497 @@
...
+For more information, see the [Implicit type conversion][03] section of this
+article.
+
+### Numeric type conversion
+
+Numeric types can be freely converted to one another as long as the target type
+is wide enough. For example:
```
mikefrobbins, line 55:
```suggestion
Numeric types can be freely converted to one another as long as the target type
can accommodate all possible values of the source type. For example:
```

The phrase "wide enough" can be vague and confusing because it doesn't clearly explain what makes one type "wider" than another.

reference/7.5/Microsoft.PowerShell.Core/About/about_Type_Conversion.md
```diff
@@ -0,0 +1,497 @@
...
+  True
+  PS> [boolean]@(1,'Hello')
+  True
+  PS> [boolean]@{}
+  True
+  ```
+
+### String type conversion
+
+A value of _any_ type can be coerced to a **String**. The default conversion is
```
mikefrobbins, line 108: **String** is capitalized and in bold in this sentence. It's in lower case and not bold in the next paragraph. Is this styling correct?

reference/7.5/Microsoft.PowerShell.Core/About/about_Type_Conversion.md
```diff
@@ -0,0 +1,497 @@
...
+[System.Reflection.TypeAttributes]::Public -bor
+    [System.Reflection.TypeAttributes]::Abstract
+```
+
+### Other type conversions
+
+A single value (non-array) can be converted to an instance of a type if:
+
+- That type has a (public) single-parameter constructor
+- And the value is same type or can be coerced to the type of the parameter
```
mikefrobbins, line 162:
```suggestion
- And the value is the same type or can be coerced to the type of the parameter
```

reference/7.5/Microsoft.PowerShell.Core/About/about_Type_Conversion.md
```diff
@@ -0,0 +1,497 @@
...
+  - In [Types.ps1xml][10] files
+  - In compiled code for types decorated with a [TypeConverterAttribute][14]
+    attribute
+  - Via classes derived from [TypeConverter][13] or [PSTypeConverter][18]
+
+### Parameter binding conversions
+
+PowerShell attempts to convert values passed to parameters to match the
+parameter type. Type conversion of parameter values occurs in cmdlets,
+functions, scripts, scriptblocks, or .NET method where the parameter is
```
mikefrobbins, line 210:
```suggestion
functions, scripts, scriptblocks, or .NET methods where the parameter is
```

reference/7.5/Microsoft.PowerShell.Core/About/about_Type_Conversion.md
```diff
@@ -0,0 +1,497 @@
...
+parameter. Parameters can also have custom conversions defined by decorating
+parameters with [ArgumentTransformationAttribute][17] attribute.
+
+For more information, see [about_Parameter_Binding][08].
+
+### Best practices for parameter binding
+
+For .NET methods, it's better to pass the exact type expected using a type
+casts where needed. Without exact types, PowerShell can select the wrong method
+overload. And, new method overloads added in future versions of .NET can break
```
mikefrobbins, line 222:
```suggestion
overload. Also, new method overloads added in future versions of .NET can break
```

reference/7.5/Microsoft.PowerShell.Core/About/about_Type_Conversion.md
```diff
@@ -0,0 +1,497 @@
...
+
+The following example shows another instance of the array conversion problem.
+
+```powershell
+PS> $bytes = [byte[]] @(1..16)
+PS> $guid = New-Object System.Guid($bytes)
+New-Object: Cannot find an overload for "Guid" and the argument count: "16".
+```
+
+PowerShell treats the `$bytes` array as a list of individual parameters. Even
```
mikefrobbins, line 283:
```suggestion
PowerShell treats the `$bytes` array as a list of individual parameters even
```

reference/7.5/Microsoft.PowerShell.Core/About/about_Type_Conversion.md
```diff
@@ -0,0 +1,497 @@
...
+though `$bytes` is an array of bytes and `System.Guid` has a `Guid(Byte[])`
+constructor.
+
+This common code pattern is instance of _pseudo method syntax_, which doesn't
```
mikefrobbins, line 287:
```suggestion
This common code pattern is an instance of _pseudo method syntax_, which doesn't
```

reference/7.5/Microsoft.PowerShell.Core/About/about_Type_Conversion.md
```diff
@@ -0,0 +1,497 @@
...
+always work as intended. This syntax translates to:
+
+```powershell
+[byte[]] $bytes = 1..16
+New-Object -TypeName System.Guid -ArgumentList $bytes # !! BROKEN
+```
+
+Given that the type of **ArgumentList** is `[object[]]`, a single argument
+that happens to be an array (of any type) binds to it _element by element_. The
+workaround is to wrap `$bytes` in an outer array so that PowerShell looks for
```
mikefrobbins, line 297:
```suggestion
workaround is to wrap `$bytes` in an outer array so that PowerShell looks for a
```

reference/7.5/Microsoft.PowerShell.Core/About/about_Type_Conversion.md
```diff
@@ -0,0 +1,497 @@
...
+
+Guid
+----
+04030201-0605-0807-090a-0b0c0d0e0f10
+```
+
+The first item of the wrapped array is our original `Byte[]` instance. That
+value matches the `Guid(Byte[]]` constructor.
+
+An alternative to array wrapping workaround is to use the intrinsic static
```
mikefrobbins, line 312:
```suggestion
An alternative to the array wrapping workaround is to use the intrinsic static
```

reference/7.5/Microsoft.PowerShell.Core/About/about_Type_Conversion.md
```diff
@@ -0,0 +1,497 @@
...
+
+```powershell
+PS> $result = [int]::MaxValue + 1
+PS> $result
+2147483648
+PS> $result.GetType().Name
+Double
+```
+
+If you want to the result to be an **Int64**, you can cast the result type or
```
mikefrobbins, line 380:
```suggestion
If you want the result to be an **Int64**, you can cast the result type or
```

reference/7.5/Microsoft.PowerShell.Core/About/about_Type_Conversion.md
```diff
@@ -0,0 +1,497 @@
...
+Int64
+PS> ([int]::MaxValue + 1L).GetType().Name
+Int64
+```
+
+The `L` suffix on the `1` literal forces the result to be an **Int64**. For
+more information, see [about_Numeric_Literals][07].
+
+Usually, it's the left-hand side (LHS) operand of PowerShell operators that
+determines the data type used in the operation. PowerShell converts (coerces)
```
mikefrobbins, line 394:
```suggestion
Usually, the left-hand side (LHS) operand of PowerShell operators determines
the data type used in the operation. PowerShell converts (coerces)
```

reference/7.5/Microsoft.PowerShell.Core/About/about_Type_Conversion.md
```diff
@@ -0,0 +1,497 @@
...
+
+For other LHS types, arithmetic operators only succeed if a given type
+custom-defines these operators via [operator overloading][02].
+
+#### Comparison operations
+
+The comparison operators, such as `-eq`, `-lt`, and `-gt`, can compare operands
+of different types. For non-strings and non-primitive types, the behavior
+depends on whether the LHS type implements interfaces such as `IEquatable` and
+`IComparable`.
```
mikefrobbins, line 461:
```suggestion
of different types. The behavior of non-strings and non-primitive types depends on
whether the LHS type implements interfaces such as `IEquatable` and `IComparable`.
```

mikefrobbins (changes requested):
See my comments.

reference/7.5/Microsoft.PowerShell.Core/About/about_Type_Conversion.md
sdwheeler: I will fix this.

PR #11623, opened 2025-01-02
Fixes #11606 - Add information about numeric parameter names

### LINKED ISSUE #11606: Add warning to avoid naming a parameter beginning with a number
### Prerequisites

### Links

https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_functions?view=powershell-7.4#functions-with-parameters
https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_functions_advanced_parameters?view=powershell-7.4

### Summary

While PowerShell allows a function parameter name to begin with a number, it should almost always be avoided.

Due to how `-` followed by a number is parsed in argument mode, using the parameter in a normal manner isn't possible. PS users should be warned to avoid this type of naming as it makes code harder to call/reuse.

The issue is _not_ apparent when writing the code/defining the command and may not be apparent to potential callers either.

### Details

```powershell
function TestFunction { param ([switch] $100, [string] $200) "100: $100"; "200: $200" }

{ TestFunction -100 }.Ast.EndBlock.Statements[0].PipelineElements[0].CommandElements[-1]
# StringConstantExpressionAst:
#     StringConstantType : BareWord
#     Value              : -100
#     StaticType         : System.String

{ TestFunction -200 Foo }.Ast.EndBlock.Statements[0].PipelineElements[0].CommandElements[1, 2]
# StringConstantExpressionAst:
#     StringConstantType : BareWord
#     Value              : -200
#     StaticType         : System.String

# StringConstantExpressionAst:
#     StringConstantType : BareWord
#     Value              : Foo
#     StaticType         : System.String
```

`-100` and `-200` are both parsed as bareword (unquoted) arguments, not parameters. This limitation is mentioned in the [Language Specification](https://learn.microsoft.com/en-us/powershell/scripting/lang-spec/chapter-02#234-parameters), albeit the `first-parameter-char` syntax suggests PowerShell outright does not support parameter names beginning with a number, which isn't accurate.

A parameter can be _named_ beginning with a number, it just can't be _used_ normally.

One workaround is hash table splatting:

```powershell
# Broken.
TestFunction -100

# 100: False
# 200: -100

# OK.
# Quoting the key isn't mandatory for this particular case, but is
# generally good practice when the intention is to specify a string.
$ht = @{ '100' = $true } 
TestFunction @ht

# 100: True
# 200:
```

Another workaround, applicable to `.ps1` script files, is to call the file with `powershell.exe -File` or `pwsh -File`, since the initial parsing of the parameter becomes irrelevant once the command line string is constructed by PS. 

### Suggested Fix

In [about_Functions](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_functions) and [about_Functions_Advanced_Parameters](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_functions_advanced_parameters):
- Add a warning that it is generally best to avoid naming a parameter beginning with a number as using the parameter normally (`Command -Parameter`) isn't possible.
  - Note that this affects any parameter-supporting command type, not just functions.
  - It applies to both advanced and non-advanced/simple commands.
- Make a note of the hash table splatting workaround in case the reader encounters someone else's code with this issue.

sdwheeler:
# PR Summary

Add information about numeric parameter names

- Fixes [AB#359338](https://dev.azure.com/msft-skilling/cebd7ef5-4282-448b-9701-88c8637581b7/_workitems/edit/359338)
- Fixes #11606

## PR Checklist

### DIFF
--- a/reference/7.5/Microsoft.PowerShell.Core/About/about_Functions_Advanced_Parameters.md
+++ b/reference/7.5/Microsoft.PowerShell.Core/About/about_Functions_Advanced_Parameters.md
@@ -1,7 +1,7 @@
 ---
 description: Explains how to add parameters to advanced functions.
 Locale: en-US
-ms.date: 07/02/2024
+ms.date: 01/02/2025
 online version: https://learn.microsoft.com/powershell/module/microsoft.powershell.core/about/about_functions_advanced_parameters?view=powershell-7.5&WT.mc_id=ps-gethelp
 schema: 2.0.0
 title: about_Functions_Advanced_Parameters
@@ -27,6 +27,60 @@ Beginning in PowerShell 3.0, you can use splatting with `@Args` to represent
 the parameters in a command. Splatting is valid on simple and advanced
 functions. For more information, see [about_Functions][14] and
 [about_Splatting][17].
+
+## Parameter declaration
+
+Parameters are variables declared in the `param()` statement of a function or
+script block. You can use the optional `[Parameter()]` attribute alone or in
+combination with the `[Alias()]` attribute or any of the parameter validation
+attributes.
+
+Parameter names follow the rules for variable names. Parameter names consist of
+decimal digits, alphabetic characters, and underscores. For a complete list of
+naming rules, see [about_Variables][20].
+
+> [!IMPORTANT]
+> It's possible to name a parameter using only decimal digits. Using numeric
+> parameter names isn't recommended because it can lead to confusion with
+> positional parameters.
+
+Consider the following example:
+
+```powershell
+function TestFunction {
+    param (
+        [switch] $100,
+        [string] $200
+    )
+
+    "100: $100"
+    "200: $200"
+}
+```
+
+If you try to use the parameters, PowerShell interprets them as negative
+numbers passed as positional parameter.
+
+```powershell
+PS> TestFunction -100 -200 Hello
+100: False
+200: -100
+$args: -200 Hello
+```
+
+The output shows that PowerShell has bound the value `-100` to the `$200`
+parameter variable. The remaining positional values are bound to `$args`. To
+work around the issue, you can use splatting to pass the parameter values.
+
+```powershell
+PS> $ht = @{100 = $true; 200 = 'Hello'}
+PS> TestFunction @ht
+100: True
+200: Hello
+$args:
+```
+
+For more information, see [about_Splatting][17].
 
 ## Type conversion of parameter values
 
@@ -98,6 +152,8 @@ Get-Date_Func: Cannot process argument transformation on parameter 'Date'.
 Cannot convert value "19-06-2018" to type "System.DateTime". Error:
 "String '19-06-2018' was not recognized as a valid DateTime."
 ```
+
+For more information, see [about_Type_Conversion](about_Type_Conversion.md).
 
 ## Static parameters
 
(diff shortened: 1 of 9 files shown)


surfingoldelephant:
Thanks for tackling this, @sdwheeler.

> It's possible to name a parameter using only decimal digits. Using numeric parameter names isn't recommended [...]

This doesn't cover the extent of the issue. The issue applies to parameter names that start with a digit, irrespective of the characters that follow.

See the following comments on the commit.

reference/7.5/Microsoft.PowerShell.Core/About/about_Functions_Advanced_Parameters.md
```diff
@@ -28,6 +28,60 @@ the parameters in a command. Splatting is valid on simple and advanced
...
+script block. You can use the optional `[Parameter()]` attribute alone or in
+combination with the `[Alias()]` attribute or any of the parameter validation
+attributes.
+
+Parameter names follow the rules for variable names. Parameter names consist of
+decimal digits, alphabetic characters, and underscores. For a complete list of
+naming rules, see [about_Variables][20].
+
+> [!IMPORTANT]
+> It's possible to name a parameter using only decimal digits. Using numeric
```
surfingoldelephant, line 43: This doesn't cover the extent of the issue.

The issue is not limited to digit-only parameter names. The issue applies to parameter names that **start with** a digit, **irrespective of the characters that follow**. 

For example, the following is just as susceptible to the raised issue, despite it _not_ using only decimal digits/being numeric.

```powershell
function TestFunction { param([string] $1Foo) "$1Foo" }
TestFunction -1Foo Bar

# Broken:
# -1Foo
```

reference/7.5/Microsoft.PowerShell.Core/About/about_Functions_Advanced_Parameters.md
```diff
@@ -28,6 +28,60 @@ the parameters in a command. Splatting is valid on simple and advanced
...
+> parameter names isn't recommended because it can lead to confusion with
+> positional parameters.
```
surfingoldelephant, line 45: This doesn't cover the extent of the issue.

The issue is not limited to digit-only parameter names. The issue applies to parameter names that **start with** a digit, **irrespective of the characters that follow**. 

For example, the following is just as susceptible to the raised issue, despite it _not_ using only decimal digits/being numeric.

```powershell
function TestFunction { param([string] $1Foo) "$1Foo" }
TestFunction -1Foo Bar

# Broken:
# -1Foo
```

reference/7.5/Microsoft.PowerShell.Core/About/about_Functions_Advanced_Parameters.md
```diff
@@ -28,6 +28,60 @@ the parameters in a command. Splatting is valid on simple and advanced
...
+        [string] $200
+    )
+
+    "100: $100"
+    "200: $200"
+}
+```
+
+If you try to use the parameters, PowerShell interprets them as negative
+numbers passed as positional parameter.
```
surfingoldelephant, line 62: This is not accurate. The token is parsed as a **string** - see the AST output in the [original example](https://github.com/MicrosoftDocs/PowerShell-Docs/issues/11606#issue-2754894126). 

To demonstrate this in practice:

```powershell
function TestFunction { param ([object] $100) $100.GetType().Name }
TestFunction -100 # String
```

reference/7.5/Microsoft.PowerShell.Core/About/about_Functions_Advanced_Parameters.md
sdwheeler: OK, I missed that distinction. I will update the docs.

surfingoldelephant: Thank you!

PR #11666, opened 2025-01-14
Lower sub-heading since it's related to previous heading

cnotin:
# PR Summary

This section about "Use the EscapeSingleQuotedStringContent() method" is only relevant in the context of the "Wrap strings in single quotes" just above so it should be a sub-heading of it instead of a same-level heading following it

See how the previous ends "A malicious user can still use single quotes in their input to inject code." and this one begins "To protect against the user inserting their own single quote characters [...]"

## PR Checklist


sdwheeler (approved):
Thanks

PR #12080, opened 2025-05-19
Fixes #12077 - Further clarification of switch behavior

### LINKED ISSUE #12077: Further clarifications regarding the `switch` statement
### Type of issue

Missing information

### Feedback

Thank you for the quick turnaround on #12067.
I'd like to suggest an amendment as well as additional improvements.

---

I've since realized that @jborean93's explanation [here](https://github.com/PowerShell/PowerShell/issues/25536#issuecomment-2873354105) implies that - possibly after parsing a bareword condition _as a number_ - _both_ the input values and non-script-block conditions are _unconditionally_ converted to strings for the comparison, which means that the equivalence between `switch` and multiple `if` statements can more accurately be expressed as follows:

```powershell
switch (<test-expression>) {
    <result1-to-be-matched> {<action>}
    <result2-to-be-matched> {<action>}
}
```

corresponds to (note the use of the subexpression operator, `$(...)`, inside `"..."`):

```powershell
if ("$(<result1-to-be-matched>)") -eq ("$(<test-expression>)") {<action>}
if ("$(<result2-to-be-matched>)") -eq ("$(<test-expression>)") {<action>}
```

The upshot is that the coerced stringification:

* _only_ yields the same results as _non-coerced_ `-eq` semantics for (a) comparing _strings to strings_ and (b) comparing _numbers to numbers_ (because of how equivalent numbers, even if specified differently, e.g. `16` vs. `0x10`, are normalized to their _decimal_ string representation)
 
* is _typically useless_ for all other types, including all non-primitive .NET types, given that the default stringification behavior of non-primitive types is to simply return the _type name_.

In other words: For guaranteed  as-is `-eq` semantics, use a _script block_ condition, `{ $_ -eq <result-to-be-matched }` (in the script block, you're free to reverse the operands as desired, for the purpose of implicit type conversions).

---

Note that the above also suggests that some of the existing framing in the "Syntax" section warrants revising:

> It's important to understand that the `<result-to-be-matched>` value is on the left-hand side of the comparison expression. That means the result of the `<test-expression>` is on the right-hand side, which can be converted to the type of the left-hand side value for comparison. 

Given the coerced, bilateral, up-front stringification, this is _not_ relevant: the order of operands is irrelevant in _string_ comparisons.

---

#### Additionally suggested improvements:

Section "Long description":

> The `switch` statement lists each condition and an optional action.

The action is _not_ optional, however.

Section "Syntax": 

> The `<test-expression>` is single expression that's evaluated in expression mode to return a value.

It may be worth introducing the concept early that if `<test-expression>` evaluates to an _enumerable_ (defined as a data type whose instances PowerShell _auto-enumerates_ in the pipeline, currently somewhat reductively called _collection_ later in the topic), the condition statements are evaluated for _each element_ resulting from the enumeration and that while for _scalar_ expressions `break` and `continue` are equivalent and exit the statement, for _enumerables_ `continue` acts differently: it continues processing, with the next enumerated element, if any.

### Page URL

https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_Switch?view=powershell-7.6

### Content source URL

https://github.com/MicrosoftDocs/PowerShell-Docs/blob/main/reference/7.6/Microsoft.PowerShell.Core/About/about_Switch.md

### Author

@sdwheeler

### Platform Id

a6a7bfc8-35a7-a768-8d1b-dcd520ee83d7

### Document Id

bfe13a02-e305-5512-42df-5cc6bcb933fb

sdwheeler:
# PR Summary

Further clarification of switch behavior. Consolidated some examples. Reorganized for better flow.

- Fixes [AB#433831](https://dev.azure.com/msft-skilling/cebd7ef5-4282-448b-9701-88c8637581b7/_workitems/edit/433831)
- Fixes #12077

## PR Checklist

### DIFF
--- a/reference/5.1/Microsoft.PowerShell.Core/About/about_Switch.md
+++ b/reference/5.1/Microsoft.PowerShell.Core/About/about_Switch.md
@@ -1,7 +1,7 @@
 ---
 description: Explains how to use a switch to handle multiple `if` statements.
 Locale: en-US
-ms.date: 05/15/2025
+ms.date: 05/19/2025
 online version: https://learn.microsoft.com/powershell/module/microsoft.powershell.core/about/about_switch?view=powershell-5.1&WT.mc_id=ps-gethelp
 schema: 2.0.0
 title: about_Switch
@@ -20,11 +20,11 @@ variables and the properties of objects.
 
 To check multiple conditions, you can use a `switch` statement. The `switch`
 statement is similar to a series of `if` statements, but it's simpler. The
-`switch` statement lists each condition and an optional action. If a condition
-matches, the action is performed.
+`switch` statement lists each condition and the corresponding action. If a
+condition matches, the action is performed.
 
-The `switch` statement can use the `$_` and `$switch` automatic variables. For
-more information, see [about_Automatic_Variables][03].
+> [!IMPORTANT]
+> The `switch` statement converts all values to strings before comparison.
 
 ## Syntax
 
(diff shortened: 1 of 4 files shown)


mklement0:
🙏

PR #12200, opened 2025-07-05
Update what-is-a-command-shell.md

yashanshh:
I was reading the documentation and found this extra provide floating


sdwheeler (approved):
@yashanshh Thanks for the fix.

PR #12481, opened 2025-11-04
using-aliases.md - fix 'mv' typo 

alirobe:
There is a typo in the documentation for move-item, where 'mi' is listed twice, and 'mv' is not mentioned. I replaced the second mention of 'mi' with 'mv'.


alirobe:
I also am seeing that the mkdir command is aliased, but the alias is not listed in this documentation. I would suggest a more thorough review, but I was just making a quick correction here.

sdwheeler:
> I also am seeing that the mkdir command is aliased, but the alias is not listed in this documentation. I would suggest a more thorough review, but I was just making a quick correction here.

`mkdir` is a function that wraps `New-Item`, not an alias.

PR #12622, opened 2025-12-30
Restructure `about_Logical_Operators` and add `-and`/`-or` precedence example

surfingoldelephant:
# PR Summary

In this PR:

- Restructures `about_Logical_Operators`, by shifting about some of the existing content and adding a new `Syntax` header.
- Adds `-and`/`-or`/`-xor` equal precedence note to `about_Logical_Operators`.
- Removes PS 6+ operators from PS 5.1 `about_Operator_Precedence` doc (in line with 5.1's `about_Operators` doc).
- Adds an `-and`/`-or` precedence example to `about_Operator_Precedence`.
- Formatting/verbiage changes.

## PR Context

The main motivation behind this PR was to highlight `-and`/`-or`'s equal precedence, which is a departure from other languages like C# and other fields like mathematics. This behavior is likely surprising to users coming from other languages/fields.

There are also other contexts _within_ PowerShell that have different operator precedence, like WQL and the AD filter (both of which give higher precedence to their AND operator). The AD filter especially is a source of confusion (users often mistakenly confuse AD filter syntax with PowerShell syntax and pass a script block to `Get-AD* -Filter`).

Based on above, I felt PowerShell's `-and`/`-or` precedence needed additional attention in both `about_Logical_Operators` and `about_Operator_Precedence`.

A real-world example where a user was surprised by PowerShell's behavior can be found [here](https://old.reddit.com/r/PowerShell/comments/1oh87gk/powershell_and_issue_with_not_very_complex_logic/).

## PR Checklist

### DIFF
--- a/reference/7.6/Microsoft.PowerShell.Core/About/about_WQL.md
+++ b/reference/7.6/Microsoft.PowerShell.Core/About/about_WQL.md
@@ -712,10 +712,8 @@ Get-CimInstance -Query $q
 - [about_Special_Characters][02]
 - [about_Quoting_Rules][01]
 - [about_WMI][04]
-- [about_WMI_Cmdlets][03]
 
 <!-- link references -->
 [01]: about_Quoting_Rules.md
 [02]: about_Special_Characters.md
-[03]: about_WMI_Cmdlets.md
 [04]: about_WMI.md


surfingoldelephant:
Regarding the invalid links above, I see now that `about_WQL.md` only exists in the 5.1 directory.

I see the `redirect_url` for `/powershell/module/microsoft.powershell.core/about/about_wql`, so is switching `about_WQL.md` to that the right approach here?

sdwheeler:
I looked into this. I think I want to add `about_WQL.md` to all versions but it needs to be updated to use CIM examples. For now, copy `about_WQL.md` to all versions. 

We will update the content in a separate PR. In the subsequent PR we need to add links to `about_WQL` to the CIM cmdlets that can use WQL.

surfingoldelephant:
> I looked into this. I think I want to add `about_WQL.md` to all versions but it needs to be updated to use CIM examples. For now, copy `about_WQL.md` to all versions.
> 
> We will update the content in a separate PR. In the subsequent PR we need to add links to `about_WQL` to the CIM cmdlets that can use WQL.

Thanks for looking into that. That change is now done.

The new invalid link warnings above are due to the new `about_WQL.md` files containing link references to v5.1-only docs.

I've created the following issue to address updating `about_WQL` and associated docs:

- https://github.com/MicrosoftDocs/PowerShell-Docs/issues/12635

sdwheeler (approved):

PR #12632, opened 2026-01-05
Define common terminology and styling

sdwheeler:
# PR Summary

Define common terminology and styling

- Fixes [AB#543257](https://dev.azure.com/msft-skilling/cebd7ef5-4282-448b-9701-88c8637581b7/_workitems/edit/543257)

## PR Checklist

### DIFF
--- a/terminology.md
+++ b/terminology.md
@@ -56,10 +56,33 @@ Example:
 Command names and parameters usually use PascalCase, but verify the correct casing by inspecting the
 command definitions.
 
-Example:
+For example, the parameters of the `TabExpansion2` function don't use PascalCase:
 
 ```powershell
-TabExpansion2 -inputScript Foo
+TabExpansion2 -inputScript "TabExpansion2 -" | Select-Object -ExpandProperty CompletionMatches
+```
+
+```Output
+CompletionText       ListItemText           ResultType ToolTip
+--------------       ------------           ---------- -------
+-inputScript         inputScript         ParameterName [string] inputScript
+-cursorColumn        cursorColumn        ParameterName [int] cursorColumn
+-ast                 ast                 ParameterName [Ast] ast
+-tokens              tokens              ParameterName [Token[]] tokens
+-positionOfCursor    positionOfCursor    ParameterName [IScriptPosition] positionOfCursor
+-options             options             ParameterName [hashtable] options
+-Verbose             Verbose             ParameterName [switch] Verbose
+-Debug               Debug               ParameterName [switch] Debug
+-ErrorAction         ErrorAction         ParameterName [ActionPreference] ErrorAction
+-WarningAction       WarningAction       ParameterName [ActionPreference] WarningAction
+-InformationAction   InformationAction   ParameterName [ActionPreference] InformationAction
+-ProgressAction      ProgressAction      ParameterName [ActionPreference] ProgressAction
+-ErrorVariable       ErrorVariable       ParameterName [string] ErrorVariable
+-WarningVariable     WarningVariable     ParameterName [string] WarningVariable
+-InformationVariable InformationVariable ParameterName [string] InformationVariable
+-OutVariable         OutVariable         ParameterName [string] OutVariable
+-OutBuffer           OutBuffer           ParameterName [int] OutBuffer
+-PipelineVariable    PipelineVariable    ParameterName [string] PipelineVariable
 ```
 
 ### Specific examples
@@ -197,7 +220,7 @@ confusion. Use the following terminology instead:
 
 ### `$_` vs. `$PSItem`
 
-`$_` and `$PSItem` are automatic varidables that refer to the current object in the pipeline.
+`$_` and `$PSItem` are automatic variables that refer to the current object in the pipeline.
 Technically, neither variable is an alias. Both variables contain the same value.
 
 `$PSItem` was added to PowerShell later in an attempt to provide a clearer purpose to the name.


sdwheeler:
@surfingoldelephant Please review and comment. We can build on this document.

terminology.md
```diff
@@ -0,0 +1,151 @@
...
+  Use the _dot-source operator_ to _dot-source_ a script.
+
+- _built-in_ drive
+- _high-level-language_ compiler
+- _member-access_ operator
+
+## Declare vs. initialize
+
+PowerShell has no concept of declaring variables, unless you count using `New-Variable` and a
+declaration. Even then the new variable it given a default value. Essentially, variables are
```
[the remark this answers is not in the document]
sdwheeler: Fixed in next suggestion.

terminology.md
```diff
@@ -0,0 +1,198 @@
...
+
+When writing about PowerShell, use the following terminology guidelines to ensure consistency across
+documentation. This is a draft document that's subject to change. Eventually, we will move this
+content to the PowerShell-Docs style guide.
+
+## Punctuation
+
+For general rules, see [Punctuation - Microsoft Style Guide][01].
+
+For PowerShell-specific terminology, see [PowerShell-Docs style guide][03]
```
surfingoldelephant:
```suggestion
For PowerShell-specific terminology, see [PowerShell-Docs style guide][03].
```

terminology.md
```diff
@@ -0,0 +1,198 @@
...
+
+- `do/until`
+- `do/while`
+- `try/catch/finally`
+- `if/elseif/else`
+
+Example:
+
+> A `do/until` loop consists of a `do` statement block followed by an `until` scriptblock
+> containing a conditional expression.
```
surfingoldelephant: `until` doesn't accept a scriptblock. The legacy language spec refers to it as the "condition" or "controlling expression".

```suggestion
> A `do/until` loop consists of a `do` statement block followed by `until` and a
> conditional expression.
```

terminology.md
```diff
@@ -0,0 +1,198 @@
...
+
+Cmdlet names and parameters usually use PascalCase, but verify the correct casing by inspecting the
+commands.
+
+### Specific examples
+
+- _filesystem_ - no space
+- _filename_ - no space
```
surfingoldelephant:
```suggestion
- _file system_
- _file name_
```

In a general sense, the Microsoft Writing Style Guide [recommends](https://learn.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/f/file-name) "file name" (and by extension, "file system").

terminology.md
```diff
@@ -0,0 +1,198 @@
...
+- _snap-in_ - the general term for a PowerShell snap-in assembly
+- _SDDL_ - abbreviation of Security Descriptor Definition Language
+- _WSMan_ and _WinRM_
+  - _WSMan_ - Microsoft's abbreviation for the WS-Management (Web Services-Management) open standard
+  - _WinRM_ - Windows Remote Management - Microsoft's implementation of the WSMan standard
+  - Use lowercase instances only where the name is lowercase in the service interfaces (such as
+    `winrm` command-line tool, or in schema URIs and other properties).
+- Variable scopes and scope modifiers
+
+  See [Appendix A - Grammar - B.1.6 Variables][04]
```
surfingoldelephant:
```suggestion
  See [Appendix A - Grammar - B.1.6 Variables][04].
```

terminology.md
```diff
@@ -0,0 +1,198 @@
...
+
+  Scope names and modifiers are capitalized:
+
+  - `Global:`
+  - `Local:`
+  - `Private:`
+  - `Script:`
+  - `Using:`
+  - `Workflow:`
```
surfingoldelephant, line 115: `Global:` pertains to the scope modifier, not the name of the scope itself.

Current docs typically (but still inconsistently) use the following format when referring to scope name:

```
To create the same variable in the **Global** scope
```

A scope modifier is used programmatically, whereas scope name is either used in a general sense or as a `-Scope` argument (e.g., `New-Variable -Scope Global`). Since arguments have bold formatting, I suggest the following:

```suggestion
  Scope names and modifiers are capitalized:

  - `Global:` and **Global** scope
  - `Local:` and **Local** scope
  - `Script:` and **Script** scope
  - `Private:`
  - `Using:`
  - `Workflow:`
```

terminology.md
```diff
@@ -0,0 +1,198 @@
...
+  ----           -----
+  COMPUTERNAME   COMPUTER01
+  ```
+
+  When identifying the environment variable namespace, use lowercase:
+
+  ```powershell
+  $env:COMPUTERNAME
+  COMPUTER01
+  ```
```
surfingoldelephant: "`Env:` modifier" isn't an accurate term here. I think it's worth drilling down into the exact terminology and functionality.

- `Env` is the name of a PowerShell drive (`PSDrive`).
- PowerShell provider cmdlets such as `Get-Item` and `Get-Content` enable access to items in a PowerShell drive.
- A PowerShell drive can be accessed by affixing the name with a `:`. An item separator (`\`) is optional if the PowerShell drive is flat (like `Env`).
- `$env:COMPUTERNAME` is an example of variable namespace notation.
- Variable namespace notation allows the items of a PowerShell drive to be accessed without explicitly calling a PowerShell provider cmdlet.
- `$env:COMPUTERNAME` is shorthand for `Get-Content -Path Env:COMPUTERNAME` and works with any PowerShell drive whose provider implements `IContentCmdletProvider`.
- Variable namespace notation requires `{}`-delineation when access of an item requires additional item separators or special characters. E.g.,  `${C:\Windows\system.ini}`, which is shorthand for `Get-Content -Path C:\Windows\system.ini`.

````suggestion
  PowerShell drive names such as `Env` are capitalized.

  ```powershell
  Get-Item -Path Env:COMPUTERNAME

  Name           Value
  ----           -----
  COMPUTERNAME   COMPUTER01
  ```

  `Env` is a special case when used in variable namespace notation.
  Use lowercase for `$env:`. For all other PowerShell drives, use normal
  capitalization.

  ```powershell
  $env:COMPUTERNAME
  COMPUTER01

  $Alias:dir
  Get-ChildItem
  ```
````

File: terminology.md [hunk 8]
```diff
@@ -0,0 +1,198 @@
...
+
+## Terms
+
+### Declare vs. initialize
+
+PowerShell has no concept of declaring variables, unless you count using `New-Variable` as a
+declaration. Even then, the new variable is given a default value. Essentially, variables are
+declared on first use. If the variable was not assigned a value, it is initialized to a default
+value depending on its type.
```
surfingoldelephant: `New-Variable -Name Foo` creates a new variable `$foo` with a value of `$null`. Are we considering `$null` as a "default value" in this case?

terminology.md
```diff
@@ -0,0 +1,198 @@
...
+
+Use _FALSE_ or _TRUE_ (all uppercase) to refer to boolean values in general writing. Use `$true` and
+`$false` (all lowercase) when referring to the boolean constants.
+
+### Hash tables
+
+- Use _hash table_ (two words, all lowercase) to refer to the computer science concept in general
+  writing. See [Hash table - Wikipedia][09].
+- Use _hashtable_ (all lowercase) when referring to the PowerShell objects of type `[hashtable]`.
+  May be capitalized as _Hashtable_ when it begins a sentence.
```
surfingoldelephant, line 206: Based on this, does `about_Hash_Tables` need to be renamed to `about_Hashtables`?

terminology.md
```diff
@@ -0,0 +1,198 @@
...
+- `[hashtable]` (with backticks) when referring to the type more specifically.
+- Full type names should match the .NET definition: `[System.Collections.Hashtable]`
+
+### Statements and script blocks
+
+A _script block_ is a generic description we have used historically, but can be a source of
+confusion. Use the following terminology instead:
+
+- A _statement block_ to refers to `{}` expressions in a statement. This is how the AST refers to
```
surfingoldelephant:
```suggestion
- A _statement block_ refers to `{}` expressions in a statement. This is how the AST refers to
```

terminology.md
```diff
@@ -0,0 +1,198 @@
...
+  them. Statement blocks don't take parameters or create new scopes.
+- _scriptblock_ should be used to refer to `{}` expressions of type `[scriptblock]`. For example,
```
surfingoldelephant, line 217: Similar to above. `about_Script_Blocks` will need to be renamed to `about_Scriptblocks`.

terminology.md
```diff
@@ -0,0 +1,198 @@
...
+
+### Native vs. external commands
+
+Not all external commands are native commands. A PowerShell script can be an _external command_, but
+it's not a native command.
+
+_Native commands_ are executables that can be run from any shell or other invocation method
+supported by the OS.
+
+### Scalar vs. single vs. singleton
```
surfingoldelephant, line 162: "Scalar" is most commonly used in reference to scalars vs collections in PowerShell. E.g., 

> "When the left-hand side operand is a scalar value, `-eq` returns a boolean value. When the operand is a collection, `-eq` acts as a filter and returns matching items from the collection. 

I think this section should specifically draw the comparison between scalar vs collection.

terminology.md
```diff
@@ -0,0 +1,198 @@
...
+- `try/catch/finally`
+- `if/elseif/else`
+
+Example:
+
+> A `do/until` loop consists of a `do` statement block followed by an `until` scriptblock
+> containing a conditional expression.
+
+Cmdlet names and parameters usually use PascalCase, but verify the correct casing by inspecting the
+commands.
```
surfingoldelephant:
````suggestion
Command names and parameters usually use PascalCase, but verify the correct casing by inspecting the
command definitions.

Example:

```powershell
TabExpansion2 -inputScript Foo
```
````

terminology.md
sdwheeler: The AST shows it as a scriptblock.
image

terminology.md
sdwheeler: Ugh. I hadn't looked into that. I don't really like it but I'll accept it.

terminology.md
surfingoldelephant: That's not valid syntax. `do { 'foo' } until ($true)` is the correct form.

terminology.md
sdwheeler: Fair point. After looking at it more closely, `New-Variable` doesn't allow you to specify a type. So, basically, this is the only way to _declare_ a variable without initialization. I will reword this.

terminology.md
sdwheeler: Yeah, that is not really accurate. It should be _singleton_ instead of _scalar_.

terminology.md
sdwheeler: Technically, yes. But practically, no. `Update-Help` doesn't have a way to rename files. If you have help already installed you end up with both files. So we can't really rename `about_` files.

terminology.md
sdwheeler: Same answer

terminology.md
sdwheeler: Oof... shows you how much I use `do/until` 😄.

PR #12832, opened 2026-03-10
docs(install): change brew install from cask to formula

GunniBusch:
# PR Summary

Powershell will be moved from homebrew cask to formula, due to the official binaries not passing gatekeeper checks. This move will fix this, but installation is a bit different. => Installation command won't require `--cask` anymore. (https://github.com/Homebrew/homebrew-core/pull/268901)

### DIFF
--- a/reference/docs-conceptual/install/install-powershell-on-macos.md
+++ b/reference/docs-conceptual/install/install-powershell-on-macos.md
@@ -205,7 +205,7 @@ brew upgrade powershell
 If you installed PowerShell with Homebrew, use the following command to uninstall:
 
 ```sh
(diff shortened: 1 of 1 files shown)


sdwheeler:
@GunniBusch Thanks for the submission, but we can't take this change now. We can revisit this if and when there is a formula-based solution.

GunniBusch:
> @GunniBusch Thanks for the submission, but we can't take this change now. We can revisit this if and when there is a formula-based solution.

Technically these instructions are also valid for the current State. Also the prs for the transition are both Reviewed and will be merge soon. See the linked pr

PR #12837, opened 2026-03-11
docs(install): change brew install from cask to formula

GunniBusch:
# PR Summary

Powershell will be moved from homebrew cask to formula, due to the official binaries not passing gatekeeper checks. This move will fix this, but installation is a bit different. => Installation command won't require `--cask` anymore. (https://github.com/Homebrew/homebrew-core/pull/268901)

The powershell formula is merged hence why I reopened this pr. Which means, that there is now a formula based solution https://github.com/Homebrew/homebrew-core/pull/268901 as stated as a requirement by 
https://github.com/MicrosoftDocs/PowerShell-Docs/pull/12832#issuecomment-4033361634

### DIFF
--- a/reference/index.yml
+++ b/reference/index.yml
@@ -11,6 +11,7 @@ metadata:
   author: sdwheeler
   ms.author: sewhee
   ms.date: 03/18/2025
+  no-loc: [Community, Discord, Slack, Spiceworks, "Stack Overflow" ]
 
 # highlightedContent section (optional)
 # Maximum of 8 items


sdwheeler:
As I said in the preview PR, we can't take this change right now. We can't support this formula-based installation. I have raised awareness of this to the engineering team. They are working on alternative solutions for installation in macOS. We can only support the solutions they provide.

GunniBusch:
@sdwheeler What do you mean? The formula based installation of honeybees is the official way of homebrew to install it. The homebrew cask is removed. You will no longer be able to install homebrew cask of powershell. And the formula is the new official way of homebrew. 

This also means that if someone would follow the current installation instructions they will get an error. Like I said. Please look at the linked pr.

sdwheeler:
The documentation has been updated. The brew instructions were moved to https://learn.microsoft.com/powershell/scripting/install/alternate-install-methods

PR #12878, opened 2026-03-23
Fixed typos in Enter-PSSession -UserName parameter description

dodexahedron:
# PR Summary

Fixed spelling "uou" -> "you" and hyphenated "key based" -> "key-based" in the [UserName parameter description of the Enter-PSSession cmdlet](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/enter-pssession#-username).

Although I also have some commentary about this parameter's description, for the part where it talks about ssh behaviors:

 * `password` and `publickey` are not the only two possibilities
 * Authentication methods are not exclusive of each other
 * There can be more than one authentication method configured.
 * There can be more than one authentication method ***required***
 * Authentication methods can be required ***in a specific sequence***.
 * `publickey`, if it appears more than once in a sequence, ***requires*** that each key be unique (ie you cannot use the same key twice)
 * (And other implementation details about openssh authentication)

The point of all that is that, in my opinion, the part of the description talking about ssh behavior seems out-of-scope for this document.

Even if it's desirable to be there for convenience, it should probably not be so absolute in its wording, at least about the sentence I fixed in this PR.

For instance, if the server has `AuthenticationMethods` set to `publickey,password` or `password,publickey` or any other sequence containing both, you *will*, in fact, be prompted for password when you have supplied a key.\
And then there are other options, like kerberos, which openssh supports and which behaves more like publickey from a user perspective, and is also not exclusive of the others.

The only invariant is that, if you supply a username, it overrides the default username presented to the server.

I'll put that into an issue if y'all prefer, and if y'all think it's worth addressing with more than a simple edit (like just deleting that text).

## PR Checklist

RTG

### DIFF
--- a/reference/7.4/Microsoft.PowerShell.Core/New-PSSession.md
+++ b/reference/7.4/Microsoft.PowerShell.Core/New-PSSession.md
@@ -2,7 +2,7 @@
 external help file: System.Management.Automation.dll-Help.xml
 Locale: en-US
 Module Name: Microsoft.PowerShell.Core
-ms.date: 12/09/2022
+ms.date: 03/23/2026
 online version: https://learn.microsoft.com/powershell/module/microsoft.powershell.core/new-pssession?view=powershell-7.4&WT.mc_id=ps-gethelp
 schema: 2.0.0
 aliases:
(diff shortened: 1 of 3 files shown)


dodexahedron:
I just took a look at the New-PSSession command too, since I figured it might have the same typos. It does not, and the text is slightly different.

Seems to me these two commands should use the same text for this parameter, whatever that ends up being.

sdwheeler:
> I'll put that into an issue if y'all prefer, and if y'all think it's worth addressing with more than a simple edit (like just deleting that text).

@dodexahedron Thanks for the edits and feedback. I changed the wording to be more general about authentication. I think this is all we need to say about it in this context.

